PR #30, opened 2023-07-06
Fix account logging

FabijanC:
## Usage related changes

- Add logging of seed and initial balance of predeployed accounts
- `GET /predeployed_accounts` was returning 0 for initial balance
- Converted initial balance to be a decimal string in output

- Rename `PredeployedAccount` to `SerializableAccount`
  - was misleading because we have both `Account` and `PredeployedAccount` and the only purpose of what was `PredeployedAccount` is being serialized (hence the renaming)
- CircleCI claims to be failing, but that's because https://github.com/0xSpaceShard/starknet-devnet-rs/pull/31 hasn't been merged.
- Format the code (might clash with #31)
- Use bigint for initial account balance
- Renamed `fn predeployed_accounts` to `fn get_predeployed_accounts`
  - should be a verb
- Move account logging into a separate function
- Add +nightly to `cargo fmt` invocation

## Checklist:

- [x] Documented the changes
- [x] Updated the tests

### DIFF
--- a/crates/starknet-server/src/api/http/models.rs
+++ b/crates/starknet-server/src/api/http/models.rs
@@ -63,8 +63,8 @@ pub(crate) struct Time {
 }
 
 #[derive(Serialize)]
-pub(crate) struct PredeployedAccount {
-    pub(crate) initial_balance: u128,
+pub(crate) struct SerializableAccount {
+    pub(crate) initial_balance: String,
     pub(crate) address: String,
     pub(crate) public_key: String,
     pub(crate) private_key: String,
@@ -73,11 +73,6 @@ pub(crate) struct PredeployedAccount {
 #[derive(Deserialize)]
 pub(crate) struct ContractAddress {
     contract_address: ContractAddressHex,
-}
-
-#[derive(Serialize)]
-pub(crate) struct ContractCode {
-    program: String,
 }
 
 #[derive(Serialize)]
--- a/crates/starknet-server/src/api/http/models.rs
+++ b/crates/starknet-server/src/api/http/models.rs
@@ -63,9 +63,9 @@ pub(crate) struct Time {
 #[derive(Serialize)]
 pub(crate) struct SerializableAccount {
     pub(crate) initial_balance: String,
-    pub(crate) address: String,
-    pub(crate) public_key: String,
-    pub(crate) private_key: String,
+    pub(crate) address: ContractAddressHex,
+    pub(crate) public_key: FeltHex,
+    pub(crate) private_key: FeltHex,
 }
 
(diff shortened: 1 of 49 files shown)


marioiordanov:
Why initial_account_balance type is changed to felt

crates/starknet-server/src/api/http/models.rs
```diff
@@ -63,8 +61,8 @@ pub(crate) struct Time {
 }
 
 #[derive(Serialize)]
-pub(crate) struct PredeployedAccount {
-    pub(crate) initial_balance: u128,
+pub(crate) struct SerializableAccount {
+    pub(crate) initial_balance: String,
```
marioiordanov, line 65: There is a type FeltHex, that was is not used for some reason in this struct. It serialises to 0x.....

crates/starknet-server/src/api/http/models.rs
```diff
@@ -63,8 +61,8 @@ pub(crate) struct Time {
 }
 
 #[derive(Serialize)]
-pub(crate) struct PredeployedAccount {
-    pub(crate) initial_balance: u128,
```
marioiordanov, line 67: In the documentation of the pythonic devnet the initial_balance is in scientific notation

crates/starknet/schemas/predeployed_accounts_fixed_seed.json
marioiordanov, line 1: As noted in previous comment in the pythonic devnet the initial balance is returned as number in scientific notation

crates/starknet-server/src/api/http/models.rs
marioiordanov: EDIT: Discussed to make it use BigInt and serialize not in scientific notation. To prevent losing precision

marioiordanov (changes requested):
Its not the best option to return in hex string the initial_balance in the Response of get_predeployed_accounts

crates/starknet-server/src/api/http/models.rs
FabijanC: I implemented the decimal string approach using bigint: https://github.com/0xSpaceShard/starknet-devnet-rs/pull/30/commits/5a756b5907c35d68270e72ad1cda795c967420ee

I introduced a new trait called `ToDecimalString` instead of using the built-in `ToString` since we already have some stringifying methods. Let me know if I should change this.

crates/starknet/schemas/predeployed_accounts_fixed_seed.json
FabijanC: We decided it should return a decimal string. Resolving.

PR #50, opened 2023-07-11
Add RPC method starknet_getClassHashAt

### LINKED ISSUE #47: get_class_hash_at

FabijanC:
## Usage related changes

- Close #47 
- If you run Devnet with `DEVNET_PORT=5050 cargo run -- --seed 0 --accounts 1`, then this works:
```shell
curl localhost:5050/rpc -X POST -H "Content-Type: application/json" --data '{
        "jsonrpc":"2.0",
        "method":"starknet_getClassHashAt",
        "params": {
            "block_id": "latest",
            "contract_address": "0x64b48806902a367c8598f4f95c305e8c1a1acba5f082d294a43793113115691"
        },
        "id":1
}'
```

- Add getter of state at specific block (currently dummy logic - only suppots latest/pending).

## Checklist:

- [x] Documented the changes
- [x] Updated the tests


crates/starknet-server/src/api/json_rpc/endpoints.rs
```diff
@@ -79,7 +80,15 @@ impl JsonRpcHandler {
         _block_id: BlockId,
         _contract_address: ContractAddressHex,
     ) -> RpcResult<ClassHashHex> {
-        Err(error::ApiError::ContractNotFound)
+        let parsed_address = _contract_address.0.try_into().unwrap(); // TODO better error handling
```
taco-paco, line 83: argument names with `_` are used for arguments that aren't used in function in rust. All arguments that are used within the function shall be without `_` in front

taco-paco, line 83:
```suggestion
        let parsed_address = _contract_address.0.try_into()?;
```

crates/starknet-server/src/api/models/mod.rs
```diff
@@ -41,6 +41,23 @@ pub enum BlockId {
     Tag(Tag),
 }
 
+impl From<BlockId> for starknet_rs_core::types::BlockId {
```
taco-paco, line 44: let's import` starknet_rs_core::types::`. If there're name collisions you can use `as`

crates/starknet-server/src/api/models/mod.rs
```diff
@@ -41,6 +41,23 @@ pub enum BlockId {
...
+    fn from(block_id: BlockId) -> Self {
+        type ImportedBlockId = starknet_rs_core::types::BlockId;
```
taco-paco, line 46: If imported can be removed

crates/starknet/src/lib.rs
```diff
@@ -5,17 +5,20 @@ use blocks::{StarknetBlock, StarknetBlocks};
...
 use starknet_api::block::{BlockNumber, BlockStatus, BlockTimestamp, GasPrice};
+use starknet_in_rust::core::errors::state_errors::StateError;
 use starknet_in_rust::definitions::block_context::{BlockContext, StarknetOsConfig};
 use starknet_in_rust::definitions::constants::{
     DEFAULT_CAIRO_RESOURCE_FEE_WEIGHTS, DEFAULT_CONTRACT_STORAGE_COMMITMENT_TREE_HEIGHT,
     DEFAULT_GLOBAL_STATE_COMMITMENT_TREE_HEIGHT, DEFAULT_INVOKE_TX_MAX_N_STEPS,
     DEFAULT_VALIDATE_MAX_N_STEPS,
 };
 use starknet_in_rust::execution::TransactionExecutionInfo;
+use starknet_in_rust::state::in_memory_state_reader::InMemoryStateReader;
```
taco-paco, line 16: Let's create `errors.rs` inside `starknet` crate and initialize there errors for the module/crate. Generally error-enum-per-crate/module pattern shall be used. 
If this error will happen one could think that the error is in `starknet_in_rust` repo

crates/starknet-server/src/api/json_rpc/endpoints.rs
```diff
@@ -79,7 +80,15 @@ impl JsonRpcHandler {
...
+
+        let starknet = self.api.starknet.read().await;
+        let state =
+            starknet.get_state_reader_at(&_block_id.into()).map_err(|_| ApiError::BlockNotFound)?;
```
ivpavici, line 87: In other places in the code it's called like this:
`error::ApiError`
So we should probably decide on how to do it consistently?

crates/starknet-server/src/api/json_rpc/endpoints.rs
FabijanC: Addressed in https://github.com/0xSpaceShard/starknet-devnet-rs/pull/50/commits/7e468c61b374117e62f877894b80f76ae2340c77

FabijanC: Wasn't as simple as that because I had to map the error. Addressed in https://github.com/0xSpaceShard/starknet-devnet-rs/pull/50/commits/7e468c61b374117e62f877894b80f76ae2340c77

crates/starknet-server/src/api/models/mod.rs
FabijanC: Addressed in https://github.com/0xSpaceShard/starknet-devnet-rs/pull/50/commits/cbc1ccbc413b5744395a68f3c289a4f412b0afcd

crates/starknet-server/src/api/json_rpc/endpoints.rs
FabijanC: Addressed in https://github.com/0xSpaceShard/starknet-devnet-rs/pull/50/commits/c14d56b31a6c8357e53efbe92519afa7d076448d

crates/starknet/src/lib.rs
FabijanC: Not sure I get the point. Here I am relying on `StateError` defined in our `crates/types/src/error.rs` and it wraps the `starknet_in_rust::core::errors::state_errors::StateError`, so won't it be clear that it's a Devnet error?

On another hand, I assume the return type of my function should be `DevnetResult` (as it already is), but that type requires the Error type to be `crate::error::Error` (the `crate` being `types`).

crates/starknet-server/src/api/json_rpc/endpoints.rs
```diff
@@ -76,10 +78,24 @@ impl JsonRpcHandler {
...
-        Err(error::ApiError::ContractNotFound)
+        let parsed_address = contract_address.0.try_into().map_err(|_| {
+            error::ApiError::RpcError(RpcError::invalid_params(format!(
+                "Invalid contract_address: {:?}",
+                contract_address.0
+            )))
+        })?;
+
+        let starknet = self.api.starknet.read().await;
+        let state = starknet
```
taco-paco, line 92: We can move this logic in `Starknet` class itself. Then it can be reused by other calls. Let's just propagate call from `JsonRpcHandler` to `Starknet`. And handle Result in `JsonRpcHandler`

taco-paco: This way we also won't need getter `.get_state_reader_at`

crates/starknet-server/src/api/json_rpc/endpoints.rs
```diff
@@ -1,3 +1,5 @@
+use server::rpc_core::error::RpcError;
+use starknet_types::felt::Felt;
```
taco-paco, line 2: this can be reshaped into

`use starknet_types::{felt::Felt, starknet_api::block::BlockNumber};`

This is preferred way it allows to save space and a general approach of imports in rust

FabijanC: Interesting, it seems our formatter prefers it the way it is (i.e. separate lines). So I'll just resolve this conversation.

crates/starknet/src/lib.rs
taco-paco: This error can appear in any crate that uses that type. That's why its important to declare errors in every crate and large modules. I created a error enum for starknet crate. It automatically implements from functions

```
#[derive(Error, Debug)]
pub enum Error {
    #[error(transparent)]
    TypesError {
        #[from]
        source: starknet_types::error::Error,
        backtrace: Backtrace,
    },
}
```

crates/starknet-server/src/api/json_rpc/endpoints.rs
taco-paco: The formatter doesn't automatically create such import so its up to developers. Otherwise our import list will get too big

FabijanC: Not sure what you mean, I tried applying the change you proposed, and after running `./scripts/format.sh`, it was back to as it is now.

FabijanC: Our CI/CD checks that all formatter rules are obeyed and I'm ok with listening to the formatter.

crates/starknet/src/lib.rs
FabijanC: >  I created a error enum for starknet crate.

Where? In your (unmerged) PR?

taco-paco: Will do in my pr

crates/starknet-server/src/api/json_rpc/endpoints.rs
taco-paco: Will create separate for 

`imports_granularity = "Module"` -> i`mports_granularity = "Crate"`

FabijanC:
Not to block further development, I'll deal with integration tests separately.

PR #56, opened 2023-07-12
Added block_number RPC endpoint

ivpavici:
## Usage related changes

Closes https://github.com/0xSpaceShard/starknet-devnet-rs/issues/45

## Checklist:

- [ ] Documented the changes
- [ ] Updated the tests


crates/starknet/src/lib.rs
```diff
@@ -355,4 +360,27 @@ mod tests {
 
         assert_eq!(block_ctx.block_info().block_number, initial_block_number + 1);
     }
+
+    #[test]
+    fn returns_block_number() {
+        let config = starknet_config_for_test();
+        let mut starknet = Starknet::new(&config).unwrap();
+
+        let mut tx = dummy_declare_transaction_v1();
```
ivpavici, line 369: This block creation logic should probably be extracted to some helper

crates/starknet/src/lib.rs
```diff
@@ -221,6 +221,11 @@ impl Starknet {
 
         Ok(())
     }
+
+    pub fn block_number(&self) -> DevnetResult<BlockNumber> {
+        let block_len = self.blocks.num_to_block.len();
```
ivpavici, line 226: not sure btw here, will this case be always >0, because of the genesis block, and do we need to handle some edge cases around that

FabijanC: This is only good under two conditions:
- there is no genesis block with number 0
- the first mined block has number 1

A more long term approach would b to get the last block and extract the number from it. And if there is no block, perhaps we can raise an error or return -1.

FabijanC (changes requested):

PR #64, opened 2023-07-13
Add getClassHashAt e2e tests; fix BackgroundDevnet

FabijanC:
## Usage related changes

None

- Fix TOCTOU error in BackgroundDevnet spawning
  - There was a race condition in obtaining a free port
- Add e2e test for RPC method starknet_getClassHashAt
  - Code duplication which I didn't know how to extract to a separate function due to some types missing which I didn't want to import/install.
- Add new dependency - lazy_static
  - Causing a warning reported in Cargo.toml - CI/CD is still passing (delegated to issue #66)

## Checklist:

- [x] Documented the changes
- [x] Updated the tests


crates/starknet-server/tests/common/mod.rs
```diff
@@ -1,53 +1,78 @@
...
+    pub const MIN_PORT: u16 = 1025;
+    pub const MAX_PORT: u16 = 65_535;
+    pub const SEED: usize = 42;
+    pub const ACCOUNTS: usize = 3;
+
+    pub const MAX_RETRIES: usize = 10;
+
+    // predeployed account info with seed=42
+    pub const PREDEPLOYED_ACCOUNT_ADDRESS: &str =
+        "0x34ba56f92265f0868c57d3fe72ecab144fc96f97954bbbc4252cef8e8a979ba";
```
mikiw, line 13: Maybe add/use constants from this file https://github.com/0xSpaceShard/starknet-devnet-rs/blob/main/crates/starknet/src/constants.rs ?

crates/starknet-server/tests/common/mod.rs
```diff
@@ -1,53 +1,78 @@
...
-    NoFreePorts,
+#[cfg(test)]
+pub mod constants {
+    pub const HOST: &str = "localhost";
+    pub const MIN_PORT: u16 = 1025;
+    pub const MAX_PORT: u16 = 65_535;
+    pub const SEED: usize = 42;
+    pub const ACCOUNTS: usize = 3;
+
+    pub const MAX_RETRIES: usize = 10;
```
mikiw, line 9: comment about MAX_RETRIES would be nice

mikiw (commented):
The rest looks fine but I recommend asking @taco-paco or @marioiordanov for a review.

crates/starknet-server/tests/common/mod.rs
FabijanC: Regarding the predeployed account address, this is the only place where it is defined, since it's specific for seed 42 (as was used in Pythonic Devnet)

crates/starknet-server/tests/common/mod.rs
```diff
@@ -1,53 +1,78 @@
...
-use hyper::{Client, StatusCode, Uri};
-use thiserror::Error;
-
-#[derive(Error, Debug)]
-pub enum TestError {
-    #[error("Cannot start Devnet")]
-    DevnetNotStartable,
-    #[error("No free ports")]
-    NoFreePorts,
+#[cfg(test)]
```
taco-paco, line 1: You don't need macro for tests in `tests`. In src its needed so they aren't compiled in `cargo build`. Folder `tests` doesn't participate in compilation

crates/starknet-server/tests/common/mod.rs
taco-paco, line 1: Can be just `tests/common.rs`. Are there plans to add more files to common folder?

crates/starknet-server/tests/common/mod.rs
```diff
@@ -1,53 +1,78 @@
...
     }
 
-    /// Ensures the background instance spawns at a free port, checks at most `MAX_RETRIES` times
-    pub(crate) async fn spawn(&mut self) -> Result<(), TestError> {
-        let free_port = get_free_port_listener().expect("No free ports");
+    impl BackgroundDevnet {
+        /// Ensures the background instance spawns at a free port, checks at most `MAX_RETRIES`
+        /// times
+        pub(crate) async fn spawn() -> Self {
+            let _mutex_guard = BACKGROUND_DEVNET_MUTEX.lock().await;
```
taco-paco, line 66:
```suggestion
            BACKGROUND_DEVNET_MUTEX.lock().await;
```

crates/starknet-server/tests/common/mod.rs
```diff
@@ -1,53 +1,78 @@
...
+        process: Child,
     }
 
-    /// Ensures the background instance spawns at a free port, checks at most `MAX_RETRIES` times
-    pub(crate) async fn spawn(&mut self) -> Result<(), TestError> {
-        let free_port = get_free_port_listener().expect("No free ports");
+    impl BackgroundDevnet {
+        /// Ensures the background instance spawns at a free port, checks at most `MAX_RETRIES`
+        /// times
+        pub(crate) async fn spawn() -> Self {
```
taco-paco, line 65: This function can fail. Lets return `Result<Self>`

crates/starknet-server/tests/common/mod.rs
```diff
@@ -1,53 +1,78 @@
...
+            let _mutex_guard = BACKGROUND_DEVNET_MUTEX.lock().await;
 
-        self.process = Some(
-            Command::new("cargo")
+            let free_port = get_free_port_listener().expect("No free ports");
+
+            let devnet_url = format!("http://{HOST}:{free_port}");
+            let devnet_rpc_url =
+                Url::parse(format!("{}/rpc", devnet_url.as_str()).as_str()).unwrap();
```
taco-paco, line 72: We can add variant in Error enum implements url::error

```
#[error("Url error")]
UrlError(#[from] url::ParseError)
```

smth like this

then it shall be possible to do just

```
let devnet_rpc_url =
                 Url::parse(format!("{}/rpc", devnet_url.as_str()).as_str())?
```

the same is for other `unwrap`/`expect`

crates/starknet-server/tests/common/mod.rs
```diff
@@ -1,53 +1,78 @@
...
     }
-    Err(TestError::NoFreePorts)
-}
 
-pub(crate) struct BackgroundDevnet {
-    client: Client<HttpConnector>,
-    process: Option<Child>,
-}
+    lazy_static! {
+        /// This is to prevent TOCTOU errors; i.e. one background devnet might find one
```
taco-paco, line 51: I think the mutex can be then locked only in `get_free_port_listener` and not for the whole `BackgroundDevnet::spawn`?

taco-paco (approved):

crates/starknet-server/tests/common/mod.rs
FabijanC: I was following the instructions from the last paragraph here: https://doc.rust-lang.org/rust-by-example/testing/integration_testing.html

> Creating the module as tests/common.rs also works, but is not recommended because the test runner will treat the file as a test crate and try to run tests inside it.

crates/starknet-server/tests/common/mod.rs
FabijanC: Addressed the account class hash constant importing in https://github.com/0xSpaceShard/starknet-devnet-rs/pull/64/commits/1266db9b1c4f60115b22f73e4554a93f4adc5da2

crates/starknet-server/tests/common/mod.rs
FabijanC: Addressed in https://github.com/0xSpaceShard/starknet-devnet-rs/pull/64/commits/b2fecb5fc1f3956b7a53e6f0375db96291b8c72f

crates/starknet-server/tests/common/mod.rs
FabijanC: That's what I first did, but it's actually pretty important to keep a reference. Even the compiler gives a warning (not an error):
```
unused `tokio::sync::MutexGuard` that must be used
if unused the Mutex will immediately unlock
```

So I'll keep it as it is, but I can add a comment.

crates/starknet-server/tests/common/mod.rs
FabijanC: It might be safer, but still not completely safe: one BackgroundDevnet can find a free port, release the mutex, and while it's trying to acquire it, another BackgroundDevnet can find the same free port. I can also add a comment about that.

crates/starknet-server/tests/common/mod.rs
FabijanC: Removed in https://github.com/0xSpaceShard/starknet-devnet-rs/pull/64/commits/8caa93c60adebb20adc618505dc33da60fa14a8a

crates/starknet-server/tests/common/mod.rs
FabijanC: Added comment in https://github.com/0xSpaceShard/starknet-devnet-rs/pull/64/commits/9e9e63b15b30a10641637f646a87fddce4c5cd08

crates/starknet-server/tests/common/mod.rs
FabijanC: Added comment in https://github.com/0xSpaceShard/starknet-devnet-rs/pull/64/commits/56701d07e723e0f0376393ee17caeda0e1f25b86

I also accidentally commited the removal of _mutex guard, but restored it in https://github.com/0xSpaceShard/starknet-devnet-rs/pull/64/commits/748e6d07f014873ed72c0c8cb02e1eeb9bf8b17f

crates/starknet-server/tests/common/mod.rs
FabijanC: Applied in https://github.com/0xSpaceShard/starknet-devnet-rs/pull/64/commits/748e6d07f014873ed72c0c8cb02e1eeb9bf8b17f

crates/starknet-server/tests/common/mod.rs
FabijanC: Applied in https://github.com/0xSpaceShard/starknet-devnet-rs/pull/64/commits/748e6d07f014873ed72c0c8cb02e1eeb9bf8b17f

crates/starknet-server/tests/common/mod.rs
taco-paco: Didn't know

PR #65, opened 2023-07-14
Error per crate/module

taco-paco:

This pr aims to create tree like error structure. Where each crate/large module contains error module that encapsulates the errors of dependant crates. Prior most of the errors where created in types/error which would make harder to track errors since the same error could appear in different crates.

Errors from subcrate can be transformed with adding corresponding enum variant, ex:

```
#[derive(Error, Debug)]
pub enum CrateError {
  #[error("Subcrate error")]
  SubcrateError(#[from] subcrate::error::Error)
}
```

## Checklist:

- [ ] Documented the changes
- [x] Updated the tests

### DIFF
--- a/crates/starknet-server/src/api/json_rpc/endpoints.rs
+++ b/crates/starknet-server/src/api/json_rpc/endpoints.rs
@@ -1,4 +1,3 @@
-use server::rpc_core::error::RpcError;
 use starknet_types::felt::Felt;
 use starknet_types::starknet_api::block::BlockNumber;
 
@@ -81,17 +80,10 @@ impl JsonRpcHandler {
         block_id: BlockId,
         contract_address: ContractAddressHex,
     ) -> RpcResult<ClassHashHex> {
-        let parsed_address = contract_address.0.try_into().map_err(|_| {
-            error::ApiError::RpcError(RpcError::invalid_params(format!(
-                "Invalid contract_address: {:?}",
-                contract_address.0
-            )))
-        })?;
+        let parsed_address = contract_address.0.try_into()?;
 
         let starknet = self.api.starknet.read().await;
-        let state = starknet
-            .get_state_reader_at(&block_id.into())
-            .map_err(|_| error::ApiError::BlockNotFound)?;
+        let state = starknet.get_state_reader_at(&block_id.into())?;
         match state.address_to_class_hash.get(&parsed_address) {
             Some(class_hash) => Ok(FeltHex(Felt::from(*class_hash))),
             None => Err(error::ApiError::ContractNotFound),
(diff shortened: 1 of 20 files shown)


crates/starknet-server/src/api/json_rpc/endpoints.rs
```diff
@@ -81,17 +80,10 @@ impl JsonRpcHandler {
         block_id: BlockId,
         contract_address: ContractAddressHex,
     ) -> RpcResult<ClassHashHex> {
-        let parsed_address = contract_address.0.try_into().map_err(|_| {
-            error::ApiError::RpcError(RpcError::invalid_params(format!(
-                "Invalid contract_address: {:?}",
```
FabijanC, line 86: Just an observation that this way we are losing the contract address from the error message, right?

crates/starknet-server/src/api/json_rpc/mod.rs
```diff
@@ -102,12 +102,32 @@ impl<T: Serialize> ToRpcResponseResult for RpcResult<T> {
                     message: err.to_string().into(),
                     data: None,
                 },
+                err @ ApiError::TypesError(_) => RpcError {
+                    // random error coe
+                    code: server::rpc_core::error::ErrorCode::ServerError(52),
```
FabijanC, line 107: There is a typo `coe` -> `code`, but more importantly, we shouldn't return random error codes

crates/starknet/src/account.rs
```diff
@@ -37,7 +37,7 @@ impl Account {
         class_hash: ClassHash,
         contract_class: ContractClass,
         fee_token_address: ContractAddress,
-    ) -> DevnetResult<Self> {
```
FabijanC, line 40: I guess we would ideally change all occurrences of `DevnetResult`? Maybe we can delegate this to a separate issue?

crates/starknet/src/error.rs
```diff
@@ -0,0 +1,26 @@
...
+    BlockIdHashUnimplementedError,
+    #[error("Specifying block by number is currently not enabled")]
+    BlockIdNumberUnimplementedError,
+    #[error("I/O error")]
+    IoError(#[from] std::io::Error),
+    #[error("Error when reading file {path}")]
+    ReadFileError { source: std::io::Error, path: String },
+}
+
+pub type Result<T, E = Error> = std::result::Result<T, E>;
```
FabijanC, line 26: How about keeping the name `DevnetResult` to avoid confusion with `std::result::Result` that might be expected when you see `Result`?

FabijanC (changes requested):

crates/starknet-server/src/api/json_rpc/endpoints.rs
taco-paco: Now not

crates/starknet/src/account.rs
taco-paco: Yeap, I would prefer to move it into error.rs

taco-paco: But to much changes for this pr then

FabijanC: Create an issue

FabijanC (approved):

Issue #51, opened 2023-07-12
multiple instances of impl of JsonRpcHandler & Starknet


taco-paco:
Right now we have multiple instances of `JsonRpcHandler` implementation around the code in `endpoints.rs` then in `json/rpc.rs` and in write_endpoints.rs. While rust allows add implementations in other files I would prefer having them in 1 file because its confusing. I saw one implementation and then saw completely other function used on `JsonRpcHandler` got puzzled for a second and then had to go all over the code looking for this implementation

taco-paco:
Renamed to: multiple instances of impl of JsonRpcHandler & Starknet

taco-paco:
I can see a problem of large single file for `Starknet` struct. As solution propose the following

we make the following structure
/starknet
---mod.rs
---add_declaration_v1.rs
---add_declaration_v2.rs
---handle_successful_transaction.rs
---....

`mod.rs` structure will contain all the function defentions but propagate a call in corresponding file with func implementation:

```
impl Starknet {
    fn add_declaration_v1(&mut self, ....) {
         add_declaration_v1_impl(self, ....);
    }

    fn add_declaration_v1(&mut self, ....) {
         add_declaration_v1_impl(self, ....);
    }

    fn add_declaration_v2(&mut self, ....) {
         handle_successful_transaction_impl(self, ....);
    }
}
```

PR #72, opened 2023-07-17
Store Sierra contracts in starknet

### LINKED ISSUE #71: Save Sierra contract class in Starknet
In [starknet_getClassAt](https://github.com/starkware-libs/starknet-specs/blob/798c29469aba76b204bf6940a8caf0d114bcc64b/api/starknet_api_openrpc.json#L394) we have to return CONTRACT_CLASS which is Sierra contract. Right now we can't fetch Sierra contract since `InMemoryStateReader` keeps only casm classes. `Starknet` struct will need to keep Sierra contracts

taco-paco:
Now `get_class`, `get_class_at` functionality can be fully added.
Merged impls of `Starknet` into 1.

- Closes #71
- Related (addresses partially) #51

## Checklist:

- [ ] Documented the changes
- [x] Updated the tests


crates/starknet/src/starknet/mod.rs
```diff
@@ -0,0 +1,450 @@
...
+            timeout: u16::default(),
+            gas_price: u64::default(),
+            chain_id: StarknetChainId::TestNet,
+        }
+    }
+}
+
+#[derive(Default)]
+pub struct Starknet {
+    pub(in crate::starknet) state: StarknetState,
```
marioiordanov, line 68: there is no need to be defined as pub(in crate::starknet) it could be pub(crate) its the same

marioiordanov (changes requested):
IMO access modifiers should be changed.

crates/starknet/src/starknet/mod.rs
taco-paco: Those fields initially were declared private by you. This is not needed anymore? I decided to keep that logic restricting access only to friend functions in `starknet` module

marioiordanov: Where are the friend functions

taco-paco: The fields with pub(in crate::starknet) are accessible only in `starknet` module not crate so those fields aren't accessible in `crate::account` module or any other module

taco-paco: `friend` is more like c++ notation that is mimicked with `pub(in some::module)`

marioiordanov: okay, that was misunderstanding about the pub(in crate::starknet).

marioiordanov (approved):

crates/starknet/src/error.rs
```diff
@@ -19,6 +19,8 @@ pub enum Error {
     IoError(#[from] std::io::Error),
     #[error("Error when reading file {path}")]
     ReadFileError { source: std::io::Error, path: String },
+    #[error("Format error")]
+    FormatError,
```
FabijanC, line 23: Where is this used? If not used, should it be deleted?

FabijanC (commented):
Left one comment and triggered the PR (by slightly modfying README.md)

crates/starknet/src/error.rs
taco-paco: Removed

FabijanC (approved):

PR #76, opened 2023-07-18
Support starknet_call

### LINKED ISSUE #69: call

FabijanC:
## Usage related changes

- Support starknet_call RPC method
- Closes #69

- Fix tests for get_class_hash_at
- Replace `get_state_reader_at` with `get_state_at` and make it a private method
- Rename `DEVNET_DEFAULT_SEED` to `DEVNET_DEFAULT_TEST_SEED` to better reflect its purpose (since there is no default seed in production)
- There was insufficient RAM on CircleCI (`error: linking with `cc` failed: exit code: 1` )so I experimented with various things, leaving these:
  - Using `--release` build
  - Using xlarge CircleCI resource class

## Checklist:

- [x] Documented the changes
- [x] Updated the tests


crates/starknet/src/starknet/mod.rs
```diff
@@ -252,16 +252,44 @@ impl Starknet {
...
+            BlockId::Tag(_) => Ok(&self.state),
             BlockId::Hash(_) => Err(Error::BlockIdHashUnimplementedError),
             BlockId::Number(_) => Err(Error::BlockIdNumberUnimplementedError),
         }
     }
 
+    pub fn get_class_hash_at(
+        &self,
+        block_id: &BlockId,
+        contract_address: &Address,
```
marioiordanov, line 266: there is already created type in types crate - ContractAddress. Ideally it should be used and inside the method converted to starknet_in_rust type Address

crates/types/src/contract_address.rs
```diff
@@ -11,6 +11,12 @@ impl ContractAddress {
     pub fn new(felt: Felt) -> DevnetResult<Self> {
         Ok(Self(PatriciaKey::new(felt)?))
     }
+
+    /// Constructs a zero address
+    pub fn zero() -> Self {
+        // using unwrap because we are sure it works for 0x0
+        Self::new(Felt::from_prefixed_hex_str("0x0").unwrap()).unwrap()
```
marioiordanov, line 18: Felt implemented from u128, that way you are going go avoid the first unwrap.
Something like this `Self::new(Felt::from(0)).unwrap()`

marioiordanov (changes requested):

crates/starknet/src/starknet/mod.rs
```diff
@@ -252,16 +252,44 @@ impl Starknet {
...
+    ) -> Result<Felt> {
+        let state = self.get_state_at(block_id)?;
+        match state.state.address_to_class_hash.get(contract_address) {
+            Some(class_hash) => Ok(Felt::from(*class_hash)),
+            None => Err(Error::ContractNotFound),
+        }
+    }
+
+    pub fn call(&self, block_id: BlockId, function_call: FunctionCall) -> Result<Vec<Felt>> {
```
marioiordanov, line 275: I see that FunctionCall is having 3 properties, each one of them is Felt which is already present in types crate and converts via From trait to Felt252, so is it going to be easier to just accept contract_address, entry_point_selector and calldata using the already defined types in the types crate and then internally convert to starknet_in_rust types. That way maybe the `From<FunctionCall> for starknet_rs_core::FunctionCall` will not be needed

FabijanC: Addressed in https://github.com/0xSpaceShard/starknet-devnet-rs/pull/76/commits/8dae79c7a3105446869ccc667596cd3f89beb0ff

crates/types/src/contract_address.rs
FabijanC: Addressed in https://github.com/0xSpaceShard/starknet-devnet-rs/pull/76/commits/744e49d02d4331571033d65b7dc6c03a8e97c68d

crates/starknet/src/starknet/mod.rs
FabijanC: Addressed in https://github.com/0xSpaceShard/starknet-devnet-rs/pull/76/commits/15ed994521c504d98291a842eb9506c60c2e464f

crates/starknet/src/starknet/mod.rs
FabijanC: Removed unused `From<FunctionCall>` in https://github.com/0xSpaceShard/starknet-devnet-rs/pull/76/commits/250eabdf049c3fb04427457cfa7a4e33463167c4

marioiordanov (approved):
approved

PR #86, opened 2023-07-20
Invoke txn

marioiordanov:
## Usage related changes

Invoke transaction logic + rpc

Changed enum variants of Transaction to be boxed, because clippy complains about them.
Added 1 new ApiError - UnsupportedVersion

## Checklist:

- [ ] Documented the changes
- [x] Updated the tests


crates/starknet/src/starknet/add_invoke_transaction.rs
```diff
@@ -0,0 +1,266 @@
...
+use crate::transactions::{StarknetTransaction, Transaction};
+
+pub fn add_invoke_transcation_v1(
+    starknet: &mut Starknet,
+    invoke_transaction: InvokeTransactionV1,
+) -> Result<TransactionHash> {
+    let state_before_txn = starknet.state.pending_state.clone();
+    let transaction_hash = invoke_transaction.generate_hash()?;
+
+    // TODO: fix value of remaining gas, because I have no idea what to put there
```
FabijanC, line 17: Looking at the code documentation:
```rust
    /// - remaining_gas: The amount of gas that the transaction disposes.
```
It sounds like it should be `max_fee / gas_price` (not sure about the rounding - probably should be integer division - i.e. floor). But on the other hand, starknet_in_rust tests don't do this anywhere - might as well be zero or INITIAL_GAS_COST.

We have to ask lambdaclass about the exact semantics.

marioiordanov: removed in [f1bf58b](https://github.com/0xSpaceShard/starknet-devnet-rs/pull/86/commits/f1bf58b1a22bfbad2c26a34394e69444cac508fd)

FabijanC: Just for the record if anyone will ever be reading this: lambdaclass says to use INITIAL_GAS_COST (we don't know why exactly it's like that, but it is).

FabijanC (approved):

PR #99, opened 2023-07-20
Add estimate_fee support

### LINKED ISSUE #81: estimate_fee

### LINKED ISSUE #109: Bug on mint
Minting was introduced in #101.

The minting transaction is created, but it is in fact rejected. I'm fixing it, but just in case anyone else is trying to mint, currently it is not possible.

FabijanC:
## Usage related changes

- Closes #81 
- Fixes #109
- Fixes wrong declare v1 hash calculation (we were doing declare v0 hash calculation)

- Refactor minting
- Add message to `ContractError` to make it more informative - this required other parts of the code (including a test) to be modified
- Integration/e2e tests left with several TODO comments - completion delegated to #128 
- Limit jobs in `cargo test` to 6: otherwise fails due to overused memory
- Update `starknet_in_rust` version
- Introduce helper functions in `tests/common.mod.rs`

## Checklist:

- [x] Documented the changes
- [x] Updated the tests


crates/starknet-server/src/api/http/endpoints/mint_token.rs
```diff
@@ -12,36 +14,44 @@ pub(crate) async fn get_fee_token() -> HttpApiResult<Json<FeeToken>> {
...
+        starknet_rs_core::utils::get_selector_from_name("balanceOf").unwrap().into();
+    let new_balance_raw = starknet.call(
+        BlockId::Tag(BlockTag::Pending),
+        erc20_address,
+        balance_selector,
+        vec![Felt::from(address)], // calldata = the query balance
+    )?;
+
+    // format balance for output - initially it is a 2-member vector (low, high)
+    assert_eq!(new_balance_raw.len(), 2);
```
marioiordanov, line 33: Why not returning some error instead of asserting.
Is there a possible situation where the return data of call method is vector of 1 element

crates/starknet-server/src/api/json_rpc/endpoints.rs
```diff
@@ -360,3 +384,76 @@ impl JsonRpcHandler {
...
+                    broadcasted_tx.common.signature.iter().map(|s| s.into()).collect(),
+                    chain_id.to_felt(),
+                    Some(broadcasted_tx.common.nonce.into()),
+                )?,
+            ))
+        }
+        BroadcastedTransaction::Declare(BroadcastedDeclareTransaction::V1(broadcasted_tx)) => {
+            let contract_class: starknet_types::contract_class::ContractClass = broadcasted_tx
+                .contract_class
+                .try_into() // TODO why does this work if the converter is in write_endpoints.rs?
```
marioiordanov, line 417: because the trait TryFrom is implemented on the DeprecatedContractClass, the same as the way you use into methods that are defined in types crate

crates/starknet-server/tests/common/mod.rs
```diff
@@ -67,6 +83,36 @@ pub mod util {
         serde_json::from_slice(&resp_body_bytes).unwrap()
     }
 
+    /// Assumes Devnet has been run with the usual account seed and returns
+    /// the signer and address of the 0th account
+    pub fn get_predeployed_account_props() -> (LocalWallet, FieldElement) {
```
marioiordanov, line 88: why not query the devnet for predeployed_accounts, similar functionality as the minting procedure, but making a GET request. btw its done in https://github.com/0xSpaceShard/starknet-devnet-rs/pull/117/files

crates/starknet-server/tests/common/mod.rs
```diff
@@ -67,6 +83,36 @@ pub mod util {
...
+    /// dummy testing value
+    pub fn get_deployable_account_signer() -> LocalWallet {
+        let new_account_private_key = "0xc248668388dbe9acdfa3bc734cc2d57a";
+        starknet_rs_signers::LocalWallet::from(starknet_rs_signers::SigningKey::from_secret_scalar(
+            FieldElement::from_hex_be(new_account_private_key).unwrap(),
+        ))
+    }
+
+    /// resolve a path relative to the crates directory
+    pub fn resolve_crates_path(relative_path: &str) -> String {
```
marioiordanov, line 105: I assume this PR will fix this hacky way of accessing test files https://github.com/0xSpaceShard/starknet-devnet-rs/issues/124

crates/starknet-server/src/api/http/endpoints/mint_token.rs
```diff
@@ -12,36 +14,44 @@ pub(crate) async fn get_fee_token() -> HttpApiResult<Json<FeeToken>> {
...
 }
 
+/// get the balance of the `address`
+fn get_balance(
+    starknet: &Starknet,
+    address: ContractAddress,
+) -> Result<BigUint, starknet_core::error::Error> {
+    let erc20_address = Felt::from_prefixed_hex_str(ERC20_CONTRACT_ADDRESS).unwrap();
+    let balance_selector =
+        starknet_rs_core::utils::get_selector_from_name("balanceOf").unwrap().into();
```
marioiordanov, line 24: here can we avoid using unwrap

crates/starknet-server/src/api/json_rpc/endpoints.rs
```diff
@@ -360,3 +384,76 @@ impl JsonRpcHandler {
         Ok(nonce)
     }
 }
+
+fn convert_broadcasted_tx(
+    broadcasted_tx: BroadcastedTransaction,
+    chain_id: StarknetChainId,
+) -> RpcResult<starknet_in_rust_tx::Transaction> {
```
marioiordanov, line 391: why didn't use the transaction from starknet-core

crates/starknet-server/src/api/json_rpc/error.rs
```diff
@@ -38,8 +38,14 @@ pub enum ApiError {
...
     #[error("Only latest/pending block is supported")]
     OnlyLatestBlock,
-    #[error("Unsupported version")]
-    UnsupportedVersion,
+    #[error("{msg}")]
+    UnsupportedAction { msg: String },
+    #[error(transparent)]
+    TransactionError(#[from] starknet_in_rust::transaction::error::TransactionError),
+    #[error(transparent)]
+    SyscallHandlerError(
```
marioiordanov, line 46: the same here

marioiordanov (commented):
comments

crates/starknet-server/tests/common/mod.rs
FabijanC: This helper function served as a fast way of not having to hardcode `CARGO_MANIFEST_DIR` everywhere.
Yes, it can be resolved with #124.

crates/starknet-server/src/api/json_rpc/endpoints.rs
FabijanC: Actually I forgot to delete this TODO because I had accepted that it's working. I didn't know that you could just define something in another file and have it accessible in another file in the same crate.

FabijanC: Removed TODO in https://github.com/0xSpaceShard/starknet-devnet-rs/pull/99/commits/a99fe208ab5c6fdea95edcb1d90354b9244a63e1

crates/starknet-server/src/api/http/endpoints/mint_token.rs
FabijanC: This is more like a sanity check during development. If this fails, it means there is a bug and the code is not ready to be committed/released. Users won't ever see this (as long as we are properly testing the code).

crates/starknet-server/tests/common/mod.rs
FabijanC: The current approach is similar to what we had in devnet-py: always (or almost always) using devnet with --seed 42, and when you need a funded account, you just take the address and the key defined in the constants. I did not rely on the endpoint for predeployed accounts because I did not think of it; I agree that it would be more stable/robust to use that approach, although somewhat slower (more complicated under the hood).

crates/starknet-server/src/api/http/endpoints/mint_token.rs
marioiordanov: It's not a constant, but a result returned from external library. How to properly test it, so its certain that the user doesn't see it

crates/starknet-server/src/api/json_rpc/error.rs
```diff
@@ -38,8 +38,14 @@ pub enum ApiError {
     InvalidContractClass,
     #[error("Only latest/pending block is supported")]
     OnlyLatestBlock,
-    #[error("Unsupported version")]
-    UnsupportedVersion,
+    #[error("{msg}")]
+    UnsupportedAction { msg: String },
+    #[error(transparent)]
+    TransactionError(#[from] starknet_in_rust::transaction::error::TransactionError),
```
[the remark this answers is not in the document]
FabijanC: I don't precisely remember, I guess it was to allow something or all of the following:
- Use the `?` operator somewhere
- Properly propagate the error without hiding the original cause

What would be a better way of doing it?

crates/starknet-server/src/api/json_rpc/error.rs
FabijanC: Resolved in https://github.com/0xSpaceShard/starknet-devnet-rs/pull/99/commits/c3f070b62ff1982a46b8f4ac42c92e2cbb1e6cf3

crates/starknet-server/src/api/json_rpc/error.rs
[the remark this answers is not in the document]
FabijanC: Resolved in https://github.com/0xSpaceShard/starknet-devnet-rs/pull/99/commits/c3f070b62ff1982a46b8f4ac42c92e2cbb1e6cf3

crates/starknet-server/src/api/json_rpc/endpoints.rs
FabijanC: Resolved in https://github.com/0xSpaceShard/starknet-devnet-rs/pull/99/commits/c3f070b62ff1982a46b8f4ac42c92e2cbb1e6cf3

crates/starknet-server/src/api/http/endpoints/mint_token.rs
FabijanC: Ended up replacing it in https://github.com/0xSpaceShard/starknet-devnet-rs/pull/99/commits/01266b4e7c0f5c017be15f792aa313121a3c9d7b

marioiordanov (approved):

PR #100, opened 2023-07-21
Added get_block_txs_count

### LINKED ISSUE #82: get_block_txs_count

ivpavici:
## Usage related changes

Closes #82 

## Checklist:

- [x] Documented the changes
- [x] Updated the tests

### DIFF
--- a/crates/starknet/src/blocks/mod.rs
+++ b/crates/starknet/src/blocks/mod.rs
@@ -3,6 +3,7 @@ use std::collections::HashMap;
 use starknet_api::block::{BlockHeader, BlockNumber, BlockStatus};
 use starknet_api::hash::{pedersen_hash_array, StarkFelt};
 use starknet_api::stark_felt;
+use starknet_rs_core::types::{BlockId, BlockTag};
 use starknet_types::felt::{BlockHash, Felt};
 use starknet_types::traits::HashProducer;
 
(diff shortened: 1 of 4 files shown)


crates/starknet/src/blocks/mod.rs
```diff
@@ -54,6 +55,17 @@ impl StarknetBlocks {
...
     }
+
+    pub fn block_number_from_block_id(&self, block_id: &BlockId) -> Option<BlockNumber> {
+        match block_id {
+            BlockId::Number(number) => Some(BlockNumber(*number)),
+
+            BlockId::Hash(hash) => self.hash_to_num.get(&Felt::from(*hash)).cloned(),
+
+            BlockId::Tag(BlockTag::Pending) => None,
+            BlockId::Tag(BlockTag::Latest) => Some(BlockNumber(self.hash_to_num.len() as u64)),
```
marioiordanov, line 66: using the length of the collection will not produce always the desired result. Because if the assigned block number doesn't equal the length of the collection the test will fail.
Possible solution:
use the last block hash in the blocks collection to retrieve the latest block number

crates/starknet/src/starknet/mod.rs
```diff
@@ -335,6 +335,17 @@ impl Starknet {
...
+
+    pub fn get_block_txs_count(&self, block_id: &BlockId) -> Result<u64> {
+        let block;
+        if let Some(block_number) = self.blocks.block_number_from_block_id(block_id) {
+            block = self.blocks.num_to_block.get(&block_number);
+        } else {
+            return Err(Error::NoBlock);
+        }
+
+        Ok(block.unwrap().get_transactions().len() as u64)
```
marioiordanov, line 347: When block_id is of type BlockId::BlockNumber, then it will use the provided number, which will fail on unwrap, because the return type of block is Option and its value will be None. 
Possible solution:
match block {
Some(x) => x.get_transactions().len(),
None => error
}

marioiordanov (changes requested):

Issue #37, opened 2023-07-07
Use contract sources, not artifacts


FabijanC:
Instead of relying on stored compilation artifacts (.json), we should keep contract sources in our repository (single source contracts). That way we won't have to manually replace the artifacts on compiler changes.

The development setup (and the CI/CD pipeline) would include smart contract compilation as a step, and in tests we would rely on this step having been executed (and artifacts present).

One would need to introduce Cairo 0 and Cairo 1 compilation (and Solidity compilation when #173 is resolved) .

FabijanC:
We thought we would need this development feature as we had it in [Devnet-py](https://github.com/0xSpaceShard/starknet-devnet/blob/7ea5a3f9a400d59979380c278adda479b6632812/.circleci/config.yml#L31), but turns out we were able to survive without it all this time.

So let's keep on storing both sources and artifacts, preferably with compilation info either in the file/folder name or in a separate compilation info or (Scarb.toml) file.

PR #107, opened 2023-07-25
Fix RPC declare V2 inner structure

marioiordanov:
Added deserialisation from string to array, because in the RPC spec for DeclareV2 the abi part is coming as string, but in the BroadcastedDeclareTransactionInput the variant with V2 ten is expecting array. So a conversion from string to array is added.
Note: now the rpc method accepts both ways of providing the abi. Either as a string or an array.
## Checklist:

- [ ] Documented the changes
- [ ] Updated the tests


crates/starknet-server/test_data/sierra_contract_class_with_abi_as_string.json
FabijanC, line 1: Don't we already have a sierra artifact?

mikiw (approved):
Thanks! Looks like now this ABI problem will disappear!

crates/starknet-server/test_data/sierra_contract_class_with_abi_as_string.json
mikiw: this will be first one

marioiordanov: on the starknet-server project there is declareV2.json, but not Sierra artefact

marioiordanov: its in the starknet project

FabijanC: Well if we don't want to reuse it and if we want to merge this ASAP, then I'll resolve this conversation and give my approval. Also related to https://github.com/0xSpaceShard/starknet-devnet-rs/issues/37

FabijanC (approved):

crates/starknet-server/test_data/sierra_contract_class_with_abi_as_string.json
marioiordanov: It couldn't be reused, because in the starknet project the cairo_1_test.json is having the abi as an array and the deserialisation tries to receive starknet_in_rust ContractClass. In the starknet-server project the situation is different, the starknet_in_rust ContractClass which is in the BroadcastedDeclareTransaction::V2 comes as a string. Which needs different deserialisation. That is the reason for having identical test-data, but not the same in this situation

PR #112, opened 2023-07-26
Add RPC chain id

mikiw:
## Checklist:

- [x] Documented the changes
- [x] Updated the tests

### DIFF
--- a/crates/starknet/src/starknet/mod.rs
+++ b/crates/starknet/src/starknet/mod.rs
@@ -440,7 +440,7 @@ mod tests {
 
     use super::Starknet;
     use crate::blocks::StarknetBlock;
-    use crate::constants::{DEVNET_DEFAULT_INITIAL_BALANCE, ERC20_CONTRACT_ADDRESS};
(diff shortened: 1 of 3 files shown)


crates/starknet/src/starknet/mod.rs
```diff
@@ -738,4 +743,14 @@ mod tests {
 
         assert_eq!(num_one_transaction.unwrap(), 1);
     }
+
+    #[test]
+    fn returns_chain_id() {
```
FabijanC, line 748: You can also add an e2e test. You would need to modify the `spawn` method of `BackgroundDevnet` to include the chain id CLI parameter

FabijanC: Actually that would require quite a lot of changes, so you might as well skip this suggestion.

marioiordanov (approved):

crates/starknet/src/starknet/mod.rs
mikiw: yes I was checking that but this PR would change to BackgroundDevnet CLI parameter PR

PR #114, opened 2023-07-27
Example PR. FeltHex and Felt united under types

taco-paco:
Migrating and uniting Models ContractClasses with Types ContractClasses classes isn't possible without doing so FeltHex. Here's an example PR of how it is going to look.

P.S pr is into my working branch so there're some warning of unused imports during compilation. In case of wanting this in main branch I can cherry-pick those changes and make pr to main branch


FabijanC (approved):
If all the tests are passing, looks like good simplification! But let's wait for @marioiordanov

marioiordanov (approved):
Looks good

PR #155, opened 2023-08-31
Central folder for test files

marioiordanov:

- Moved all test files to test_data folder in the root of the project. Each crate has folder as soft link to the test_data in the root. Every project have access to the same files.
- Organised the test contracts in folder cairo0 and cairo 1

## Checklist:

- [x] Documented the changes
- [x] Updated the tests


taco-paco (changes requested):
Highly against this, especially with symlink which is already hardocoding so our repo compiles without a lot of changes. Types crate shall not be aware of test data of starknet crate or starknet-server crate. It breaks modularity and responsibility of a crate.

taco-paco:
> Highly against this, especially with symlink which is already hardocoding so our repo compiles without a lot of changes. Types crate shall not be aware of test data of starknet crate or starknet-server crate. It breaks modularity and responsibility of a crate.

If any test data to be reused by other create by higher crate necessary utilities can be created in child test modules.

FabijanC (commented):
Just to say that the PR looks good in the sense that it seems to implement what it states in the description. But let's reach a conclusion on if we need this change. Personally I don't see the symlinks and a common separate directory a problem. Perhaps we can create a new crate for holding test data (and potentially put e2e tests there, too).

mikiw (commented):
Because of symlink usage, it doesn't work on windows.

There is a possibility to create a script that converts Unix symlink to Windows symlink and runs it on build or other trigger but the question is it worth it? Maybe it's just easier to go with Edwin's approach now.

marioiordanov:
Converting to draft, because symlinks doesn't work on Windows as @mikiw tested.
Will have to find another solution

Issue #34, opened 2023-07-07
Exiting on invalid CLI argument value needs to be improved


FabijanC:
The PR https://github.com/0xSpaceShard/starknet-devnet-rs/pull/30 introduced a non-user-friendly exiting on invalid `--initial-balance`. This should be improved. I left a `// TODO` comment for that.

Basically anything that relies on `panic!` will not look nice on error.

Potential solution: use std::process:exit
Better solution, which might allow for the cool exiting done by clap: implement FromStr

EDIT: this now applies to several other CLI arguments, not all necessarily marked with TODO (--chain-id and --host for sure)

PR #159, opened 2023-09-07
Dump and load

mikiw:
## Checklist:

- [x] Documented the changes
- [x] Updated the tests


mikiw:
Seems that the current code will work. I'm starting to write integration tests.

mikiw:
Does this PR should include working endpoints or just the CLI feature?

FabijanC:
@mikiw I think separation into multiple PRs is good in this case.

crates/starknet-server/src/cli.rs
```diff
@@ -93,6 +105,12 @@ impl Args {
                 "TESTNET2" => StarknetChainId::TestNet2,
                 _ => panic!("Invalid value for chain-id"),
             },
+            dump_on: match self.dump_on.clone().unwrap_or_default().as_str() {
```
taco-paco, line 108: If dump_on specified shalln't we check that `dump_path != None`? What about other way around? @FabijanC

crates/starknet-server/src/main.rs
```diff
@@ -57,7 +64,22 @@ async fn main() -> Result<(), anyhow::Error> {
...
+            let mut v: Vec<u8> = Vec::new();
+            file.read_to_end(&mut v).expect("Failed to read from file");
+            let decoded: Option<String> =
+                bincode::deserialize(&v[..]).expect("Failed to deserialize starknet transactions");
+            transactions = serde_json::from_str(decoded.unwrap().as_str())
+                .expect("Failed to decode starknet transactions");
+        }
+    }
+
+    let api = api::Api::new(Starknet::new(&starknet_config, Some(transactions))?);
```
taco-paco, line 82: Isn't that always some even though `transactions` can be empty?

crates/starknet-server/src/cli.rs
FabijanC: In devnet-py, there is this check:
```python
    if parsed_args.dump_on and not parsed_args.dump_path:
        sys.exit("Error: --dump-path required if --dump-on present")
```

crates/starknet-server/src/main.rs
```diff
@@ -57,7 +64,22 @@ async fn main() -> Result<(), anyhow::Error> {
         IpAddr::from_str(starknet_config.host.as_str()).expect("Invalid value for host IP address");
     let mut addr = SocketAddr::new(host, starknet_config.port);
 
-    let api = api::Api::new(Starknet::new(&starknet_config)?);
+    // Load starknet transactions
+    let mut transactions: StarknetTransactions = StarknetTransactions::default();
+    if let Some(path) = &starknet_config.dump_path {
```
taco-paco, line 69:
```suggestion
    let transactions: Option<StarknetTransactions> = if let Some(path) = &starknet_config.dump_path {
```

taco-paco: For else branch just return None

crates/starknet-server/src/main.rs
```diff
@@ -57,7 +64,22 @@ async fn main() -> Result<(), anyhow::Error> {
         IpAddr::from_str(starknet_config.host.as_str()).expect("Invalid value for host IP address");
     let mut addr = SocketAddr::new(host, starknet_config.port);
 
-    let api = api::Api::new(Starknet::new(&starknet_config)?);
+    // Load starknet transactions
+    let mut transactions: StarknetTransactions = StarknetTransactions::default();
```
taco-paco, line 68: Let's move this to separate function

crates/starknet-server/src/main.rs
```diff
@@ -57,7 +64,22 @@ async fn main() -> Result<(), anyhow::Error> {
...
+    if let Some(path) = &starknet_config.dump_path {
+        let file_path = Path::new(path);
+        if file_path.exists() {
+            let mut file = File::open(file_path).expect("Failed to open file");
+            let mut v: Vec<u8> = Vec::new();
+            file.read_to_end(&mut v).expect("Failed to read from file");
+            let decoded: Option<String> =
+                bincode::deserialize(&v[..]).expect("Failed to deserialize starknet transactions");
```
taco-paco, line 76:
```suggestion
                bincode::deserialize(&v).expect("Failed to deserialize starknet transactions");
```

This doesn't work?

crates/starknet-server/src/main.rs
```diff
@@ -57,7 +64,22 @@ async fn main() -> Result<(), anyhow::Error> {
         IpAddr::from_str(starknet_config.host.as_str()).expect("Invalid value for host IP address");
     let mut addr = SocketAddr::new(host, starknet_config.port);
 
-    let api = api::Api::new(Starknet::new(&starknet_config)?);
+    // Load starknet transactions
+    let mut transactions: StarknetTransactions = StarknetTransactions::default();
+    if let Some(path) = &starknet_config.dump_path {
+        let file_path = Path::new(path);
+        if file_path.exists() {
```
taco-paco, line 71: There shall be else with an error I guess

crates/starknet-server/src/cli.rs
taco-paco: What if `dump_path` specified shall be dump_on specified as well?

crates/starknet/src/starknet/mod.rs
```diff
@@ -159,6 +175,95 @@ impl Starknet {
...
+                match transaction.inner.clone() {
+                    Transaction::Declare(DeclareTransaction::Version0(_tx)) => {
+                        panic!("DeclareTransactionV0V1 is not supported");
+                    }
+                    Transaction::Declare(DeclareTransaction::Version1(tx)) => {
+                        let contract_class = this
+                            .state
+                            .contract_classes
+                            .get(&tx.class_hash)
+                            .expect("Failed to load Cairo0ContractClass from state");
```
taco-paco, line 191: Please create corresponding errors in errors.rs module. This code can't use expect/panics

taco-paco: Also for all cases below

crates/starknet/src/starknet/mod.rs
```diff
@@ -159,6 +175,95 @@ impl Starknet {
 
         this.restart_pending_block()?;
 
+        // Re-execute transactions
+        if transactions.is_some() {
```
taco-paco, line 179: This definitely has to be different function

crates/starknet/src/starknet/mod.rs
```diff
@@ -159,6 +175,95 @@ impl Starknet {
...
+            // is_some() is need here?
+            for (_hash, transaction) in transactions.unwrap_or_default().iter() {
```
taco-paco, line 181: In case _hash isn't used 

```suggestion
            for (_, transaction) in transactions.unwrap_or_default().iter() {
```

crates/starknet/src/starknet/mod.rs
```diff
@@ -159,6 +175,95 @@ impl Starknet {
...
+                match transaction.inner.clone() {
+                    Transaction::Declare(DeclareTransaction::Version0(_tx)) => {
```
taco-paco, line 183:
```suggestion
                    Transaction::Declare(DeclareTransaction::Version0(_)) => {
```

crates/starknet/src/starknet/mod.rs
```diff
@@ -159,6 +175,95 @@ impl Starknet {
...
+                        let deploy_account_tx = BroadcastedDeployAccountTransaction::new(
+                            &tx.constructor_calldata,
+                            tx.max_fee,
+                            &tx.signature,
+                            tx.nonce,
+                            tx.class_hash,
+                            tx.contract_address_salt,
+                            tx.version,
+                        );
+                        println!("DeployAccount!!!!");
```
taco-paco, line 234: Looks like shall be removed

crates/starknet/src/starknet/mod.rs
```diff
@@ -159,6 +175,95 @@ impl Starknet {
...
+                        println!("DeployAccount!!!!");
+                        println!("DeployAccount!!!!");
+
+                        let result = this.add_deploy_account_transaction(deploy_account_tx);
+                        if result.is_err() {
+                            panic!("Failed to add BroadcastedDeployAccountTransaction");
+                        }
+                    }
+                    Transaction::Deploy(_tx) => {
```
taco-paco, line 243:
```suggestion
                    Transaction::Deploy(_) => {
```

crates/starknet/src/starknet/mod.rs
```diff
@@ -159,6 +175,95 @@ impl Starknet {
...
+                        panic!("DeployTransaction is not supported");
+                    }
+                    Transaction::Invoke(InvokeTransaction::Version0(_tx)) => {
```
taco-paco, line 246:
```suggestion
                    Transaction::Invoke(InvokeTransaction::Version0(_)) => {
```

crates/starknet-server/src/cli.rs
FabijanC: In devnet-py, there are two modes of dumping:
1. dumping on user's HTTP request
2. dumping on an event:
    a. after each tx
    b. on devnet exit

In that sense, providing only dump_path is ok for devnet-py, because that means the user will dump via HTTP endpoint. But I haven't reviewed this PR yet, so I don't know if the same logic applies here.

crates/starknet/src/starknet/mod.rs
```diff
@@ -467,6 +576,23 @@ impl Starknet {
         add_invoke_transaction::add_invoke_transcation_v1(self, invoke_transaction)
     }
 
+    // save starknet transactions to file
+    pub fn dump_transactions(&self) {
```
taco-paco, line 580:
```suggestion
    pub fn dump_transactions(&self) -> Result<()> {
```
Exoects have to go :(

crates/starknet/src/starknet/mod.rs
```diff
@@ -231,6 +336,10 @@ impl Starknet {
         // clear pending block information
         self.generate_pending_block()?;
 
+        if self.config.dump_on.is_some() && self.config.dump_on == Some(DumpMode::OnTransaction) {
```
taco-paco, line 339:
```suggestion
        if self.config.dump_on == Some(DumpMode::OnTransaction) {
```

crates/starknet-server/src/main.rs
```diff
@@ -77,7 +99,19 @@ async fn main() -> Result<(), anyhow::Error> {
...
 }
+
+pub async fn shutdown_signal(api: Api) {
+    tokio::signal::ctrl_c().await.expect("Failed to install CTRL+C signal handler");
+
+    // Wait for the CTRL+C signal
+    signal::ctrl_c().await.expect("Failed to read CTRL+C signal");
+
+    let starknet = api.starknet.read().await;
+    if starknet.config.dump_on.is_some() && starknet.config.dump_on == Some(DumpMode::OnExit) {
```
taco-paco, line 114:
```suggestion
    if starknet.config.dump_on == Some(DumpMode::OnExit) {
```

crates/starknet/src/starknet/mod.rs
```diff
@@ -467,6 +576,23 @@ impl Starknet {
...
+            Some(path) => {
+                let starknet_dump = Some(
+                    serde_json::to_string(&self.transactions)
+                        .expect("Failed to serialize starknet transactions"),
+                );
+                let encoded: Vec<u8> = bincode::serialize(&starknet_dump)
+                    .expect("Failed to encode starknet transactions");
+                fs::write(Path::new(&path), encoded).expect("Failed to save starknet transactions");
+            }
+            None => {} // TODO: Error here?
```
taco-paco, line 592: Yes, shall return an error. I we're here that dump action exist

crates/starknet/src/starknet/mod.rs
```diff
@@ -231,6 +336,10 @@ impl Starknet {
         // clear pending block information
         self.generate_pending_block()?;
 
+        if self.config.dump_on.is_some() && self.config.dump_on == Some(DumpMode::OnTransaction) {
+            self.dump_transactions();
```
taco-paco, line 340: Won't we have duplicate transactions with this? Can't see any logic that tracks what transactions were already dumped

crates/starknet/src/starknet/mod.rs
```diff
@@ -467,6 +576,23 @@ impl Starknet {
...
+        // TODO: attach in OnTransaction mode or dump all??
+        match &self.config.dump_path {
+            Some(path) => {
+                let starknet_dump = Some(
+                    serde_json::to_string(&self.transactions)
+                        .expect("Failed to serialize starknet transactions"),
+                );
+                let encoded: Vec<u8> = bincode::serialize(&starknet_dump)
+                    .expect("Failed to encode starknet transactions");
+                fs::write(Path::new(&path), encoded).expect("Failed to save starknet transactions");
```
taco-paco, line 590: This will override the whole file, need to use append option. Even if DumpOn::OnTransaction was activated all data in the file will be overridden

crates/starknet/src/starknet/mod.rs
```diff
@@ -104,7 +117,10 @@ pub struct Starknet {
 }
 
 impl Starknet {
-    pub fn new(config: &StarknetConfig) -> DevnetResult<Self> {
+    pub fn new(
+        config: &StarknetConfig,
+        transactions: Option<StarknetTransactions>,
```
taco-paco, line 122: IMO loading transaction shall be done by outside entity like `DumpLoader`. `Starknet` just propagates calls to starknet_in_rust and filesystem(now) loading itself looks like not his responsibility.
DumpLoader has function `load` that returns `Starknet`.

crates/starknet/src/starknet/mod.rs
taco-paco: I would propose considering creating `Dumper` entity that will have a queue of transaction hashes and when `dump_transactions` called exist queue is dumped

crates/starknet/src/starknet/mod.rs
```diff
@@ -418,6 +527,23 @@ impl Starknet {
         add_invoke_transaction::add_invoke_transaction(self, invoke_transaction)
     }
 
+    // save starknet transactions to file
+    pub fn dump_transactions(&self) {
+        // TODO: attach in OnTransaction mode or dump all??
```
mikiw, line 528: @FabijanC Fabio in the old devnet we used the same function for the dump on the transaction and on exit, so I left this comment. There was a reason to do so?

taco-paco: Also as follow up: what is the point of having `OnTransaction`  if we redump all of them?

FabijanC: Regarding why the same dumping function was used for on-transaction and on-exit dumping: I did it because it was the simplest way of dumping. I agree that appending transactions would be more efficient, but we probably won't be doing such big interventions in the old devnet.

mikiw:
> @mikiw I think separation into multiple PRs is good in this case.

endpoints will be solved here:
https://github.com/0xSpaceShard/starknet-devnet-rs/issues/174

crates/starknet/src/starknet/mod.rs
mikiw: This will be solved in separate issue

mikiw: @taco-paco and about "OnTransaction if we redump all of them" -> one of the reasons is an unexpected crash, in that case gently showdown might not be detected and data can be lost

crates/starknet/src/starknet/mod.rs
mikiw: The current implementation is just dumping them all into a file on every new transaction.

After a daily talk today, we decided to handle this optimization in a new issue https://github.com/0xSpaceShard/starknet-devnet-rs/issues/175 

Also is still not clear if re-execution will be the final implementation of dump and load.

crates/starknet/src/starknet/mod.rs
mikiw: Removed in d6ceb9f587a58f47653d937c1883e6fe3d172f4f

crates/starknet-server/src/main.rs
mikiw: Yes! Removed in 576d264dcbff6e92b1f95840b783cc3b5637cbf4

crates/starknet/src/starknet/mod.rs
mikiw: fixed in 36c55738b8ce64d93912e10c4780b1f2424f1e67 is that ok?

crates/starknet-server/src/main.rs
mikiw: Fixed in 3028e91e1f57b71f5d9daae23a39ce7bfbce8b90

crates/starknet-server/src/main.rs
mikiw: Fixed in 3028e91e1f57b71f5d9daae23a39ce7bfbce8b90

crates/starknet-server/src/main.rs
mikiw: Fixed in 3028e91e1f57b71f5d9daae23a39ce7bfbce8b90

crates/starknet-server/src/main.rs
mikiw: Fixed in 3028e91e1f57b71f5d9daae23a39ce7bfbce8b90

crates/starknet/src/starknet/mod.rs
mikiw: fixed in 6eae5a26d2a3fed5d02bcbab038cc9c232d227b3

crates/starknet/src/starknet/mod.rs
mikiw: fixed in aca53690763aa8ccba3a03b49830d8cc260ae093

crates/starknet/src/starknet/mod.rs
mikiw: fixed in aca53690763aa8ccba3a03b49830d8cc260ae093

crates/starknet/src/starknet/mod.rs
mikiw: fixed in aca53690763aa8ccba3a03b49830d8cc260ae093

crates/starknet-server/tests/test_dump_and_load.rs
```diff
@@ -0,0 +1,272 @@
...
+            json!({
+                "address": DUMMY_ADDRESS,
+                "amount": DUMMY_AMOUNT
+            })
+            .to_string(),
+        );
+        let resp = devnet_dump.post_json("/mint".into(), req_body).await.unwrap();
+        assert_eq!(resp.status(), StatusCode::OK, "Checking status of {resp:?}");
+        let mut resp_body = get_json_body(resp).await;
+        let tx_hash_value = resp_body["tx_hash"].take();
```
FabijanC, line 55: No need to use `.take()`, here nor in the other place where it's used. `take` leaves a `Null` in the place of the taken value. What I would recommend is to use `resp_body["tx_hash"].as_str().unwrap()`, because it will also allow to replace the horribly long `FieldElement::from_hex_be(tx_hash_value.as_str().unwrap()).unwrap()` with a simpler `FieldElement::from_hex_be(tx_hash_value).unwrap()`.

FabijanC (changes requested):

crates/starknet/src/starknet/mod.rs
```diff
@@ -419,6 +525,28 @@ impl Starknet {
...
 
+    /// save starknet transactions to file
+    pub fn dump_transactions(&self) -> DevnetResult<()> {
+        match &self.config.dump_path {
+            Some(path) => {
+                let starknet_dump = Some(
+                    serde_json::to_string(&self.transactions)
+                        .expect("Failed to serialize starknet transactions"),
+                );
+                let encoded: Vec<u8> = bincode::serialize(&starknet_dump)
```
taco-paco, line 536: Expect

taco-paco (changes requested):
In my opinion `dump_transactions` code is not production ready. Rewriting all transactions in file will turn into terrible user experience. I see that the new issue is opened but this pr is definitely not ready for main even if all comments are fixed. New pr with optimisations can be opened into this one

crates/starknet/src/starknet/mod.rs
taco-paco: Don't see this problem. Data shall be flushed on every dump call. So if we unexpectedly crush we will get the same state as with current way

crates/starknet-server/src/main.rs
taco-paco: Added a comment in commit

crates/starknet/src/starknet/mod.rs
taco-paco: Discussed in dm.

crates/starknet/src/starknet/mod.rs
taco-paco: There's [SerializeSeq trait](https://docs.rs/serde/latest/serde/ser/trait.SerializeSeq.html) that can help with that. Some modifications has to made tho to always get valid json

crates/starknet-server/tests/test_dump_and_load.rs
```diff
@@ -0,0 +1,272 @@
...
+
+        // Although dump and load is a multiplatform feature and works on all systems is troublesome
```
taco-paco, line 117: Tests can't become platform dependent. Utilise     #[cfg(windows)] for creating 2nd implementation for killing process on windows

taco-paco: From chatgpt.

```
    #[cfg(unix)]
    {
        Command::new("kill")
            .args(&[process_id.to_string(), signal])
            .status()
    }

    #[cfg(windows)]
    {
        // On Windows, you can use a third-party library like `ctrlc` to send signals.
        // Add `ctrlc` as a dependency in your Cargo.toml.
        // https://crates.io/crates/ctrlc
        let _ = ctrlc::send_signal(process_id, ctrlc::Signal::INT);
        Ok(ExitStatus::from_exit_code(0)) // Return a dummy exit status.
    }
```

taco-paco: Also a mock entrypoint can be introduced under separate testing feature. Other cases will have to be resolved via support

crates/types/src/rpc/transactions/declare_transaction_v2.rs
```diff
@@ -9,10 +10,12 @@ use crate::felt::{
     TransactionVersion,
 };
 
-#[derive(Debug, Clone, Default, Eq, PartialEq, Deserialize, Serialize)]
+#[derive(Debug, Clone, Eq, PartialEq, Deserialize, Serialize)]
+#[serde(deny_unknown_fields)]
 pub struct DeclareTransactionV2 {
     pub class_hash: ClassHash,
     pub compiled_class_hash: CompiledClassHash,
+    pub contract_class: SierraContractClass,
```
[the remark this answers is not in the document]
mikiw: yes

crates/starknet/src/transactions.rs
```diff
@@ -44,12 +48,14 @@ impl HashIdentified for StarknetTransactions {
...
 
 #[allow(unused)]
+#[derive(Debug, Serialize, Deserialize)]
 pub struct StarknetTransaction {
     pub inner: Transaction,
     pub(crate) finality_status: Option<TransactionFinalityStatus>,
     pub(crate) execution_result: ExecutionResult,
     pub(crate) block_hash: Option<BlockHash>,
     pub(crate) block_number: Option<BlockNumber>,
+    #[serde(skip_serializing, skip_deserializing)]
```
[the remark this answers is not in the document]
mikiw: 1. It will be regenerated
2. TransactionExecutionInfo is not Serialize, Deserialize in starknet in rust (maybe it will be in future)

README.md
```diff
@@ -121,6 +121,33 @@ Devnet predeploys a [UDC](https://docs.openzeppelin.com/contracts-cairo/0.6.1/ud
...
+
+### Loading
+
+To load a preserved Devnet instance, the options are:
+- Loading on startup:
+```
+cargo run -- --load-path <PATH>
+```
+
+Current load implementation is re-executing transactions stored in the dump file so the state of the blockchain can differ. For example, the timestamps of `StarknetBlock` will be different.
```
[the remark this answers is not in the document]
mikiw: True, it's more intuitive, fixed in 68a2a88243793ec097ebade098defbc2d73b9450, and about time-dependent functionality there is not much we can do about this now.

I remember that in python devnet we had some time manipulation features, maybe after a couple of months it will be introduced in devnet in rust, or maybe in the future, we will move from re-execution to dump state.

README.md
```diff
@@ -121,6 +121,33 @@ Devnet predeploys a [UDC](https://docs.openzeppelin.com/contracts-cairo/0.6.1/ud
...
 
+## Dumping & Loading
+
+To preserve your Devnet instance for future use, there are several options:
+- Dumping on exit (handles Ctrl+C, i.e. SIGINT, doesn't handle SIGKILL):
+```
+cargo run -- --dump-on exit --dump-path <PATH>
+```
+
+- Dumping after each transaction (done in background, doesn't block):
```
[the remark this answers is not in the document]
mikiw: removed in 68a2a88243793ec097ebade098defbc2d73b9450

README.md
```diff
@@ -121,6 +121,33 @@ Devnet predeploys a [UDC](https://docs.openzeppelin.com/contracts-cairo/0.6.1/ud
...
+- Loading on startup:
+```
+cargo run -- --load-path <PATH>
+```
+
+Current load implementation is re-executing transactions stored in the dump file so the state of the blockchain can differ. For example, the timestamps of `StarknetBlock` will be different.
+
+### Cross-version disclaimer
+
+Dumping and loading is not guaranteed to work cross-version. I.e. if you dumped one version of Devnet (e.g. 0.1.0), do not expect it to be loadable with a different version (e.g. 0.2.0).
```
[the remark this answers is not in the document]
mikiw: removed in 68a2a88243793ec097ebade098defbc2d73b9450

README.md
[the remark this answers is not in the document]
FabijanC: yes, maybe we can add to docs this caveat note about time being changed

crates/starknet-server/src/main.rs
mikiw: Fixed in 2e2312e9305304bcc68c0c2858ff094e6edf5a91

README.md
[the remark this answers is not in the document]
mikiw: but in the version that you suggested, it's self-explanatory:
"Conversely, loading is implemented as the re-execution of transactions from a dump. "

crates/starknet-server/tests/test_dump_and_load.rs
```diff
@@ -0,0 +1,272 @@
...
+            .unwrap();
+
+        let predeployed_account = Arc::new(predeployed_account);
+
+        // deploy the contract
+        let contract_factory =
+            ContractFactory::new(declaration_result.class_hash, predeployed_account.clone());
+        let deploy_result = contract_factory
+            .deploy(vec![], FieldElement::ZERO, false)
+            .max_fee(FieldElement::from(1000000000000000000000000000u128))
```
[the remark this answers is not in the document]
mikiw: Fixed in d2ecda030209335602b0a0a0c2ff81bd61b64d18

crates/starknet/src/starknet/mod.rs
mikiw: Fixed in 27ec8cc63261fb9430c873a0adda2ceebf38f005

README.md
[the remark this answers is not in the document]
FabijanC: Ok, you don't have to modify it. But if users come complaining about their timestamp logic not working, I'll tag you to explain :stuck_out_tongue:

crates/starknet-server/src/main.rs
mikiw: responded in commit

crates/starknet/src/starknet/mod.rs
mikiw: Fixed in 27ec8cc63261fb9430c873a0adda2ceebf38f005 but it will be refactored later as separate module

crates/starknet/src/starknet/mod.rs
mikiw: Fixed in 27ec8cc63261fb9430c873a0adda2ceebf38f005

crates/starknet/src/starknet/mod.rs
mikiw: Fixed in 2e2312e9305304bcc68c0c2858ff094e6edf5a91

crates/starknet/src/starknet/mod.rs
mikiw: Fixed in 27ec8cc63261fb9430c873a0adda2ceebf38f005

crates/starknet/src/starknet/mod.rs
mikiw: The new update is that even with blockier we will do re-execution optimization can be done in this PR.

crates/starknet/src/starknet/mod.rs
mikiw: I'll call you or dm about that.

crates/starknet/src/starknet/mod.rs
mikiw: Now transactions are removed in the constructor but I'll call you or dm about DumpLoader.

crates/starknet-server/src/cli.rs
mikiw: Fixed in 886ddf0bdaef2dd5e9a2470b1e24d782e2baf6fe and 855d916694823aa429a1c65bde11740e9dc4f7ff

FabijanC: Marking this conversation as resolved, but I guess we'll deal with these unfriendly panics in the same PR that will resolve #34.

mikiw:
> In my opinion `dump_transactions` code is not production ready. Rewriting all transactions in file will turn into terrible user experience. I see that the new issue is opened but this pr is definitely not ready for main even if all comments are fixed. New pr with optimisations can be opened into this one

The situation with blockifier is clarified so we can optimize this logic in this PR.

crates/starknet-server/src/main.rs
```diff
@@ -75,7 +97,19 @@ async fn main() -> Result<(), anyhow::Error> {
     info!("Starknet Devnet listening on {}", addr);
 
     // spawn the server on a new task
-    let serve = tokio::task::spawn(server);
+    let serve = tokio::task::spawn(server.with_graceful_shutdown(shutdown_signal(api.clone())));
 
     Ok(serve.await??)
 }
+
+pub async fn shutdown_signal(api: Api) {
```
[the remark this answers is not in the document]
mikiw: I liked the second option more, fixed in ef49525891d6f0a658d697a8c3fc8162806565ba

README.md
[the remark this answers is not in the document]
mikiw: It's quick so I added this in 5a8c50aeca496656f0aa7c7baf25769b22282808

Cargo.lock
```diff
@@ -1587,7 +1607,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9f4af8c3e7be5ac46b7da4a3ab551ee4c8c8e2fdb501a4dda4dceef4c66dbcde"
 dependencies = [
  "anyhow",
- "bincode",
+ "bincode 2.0.0-rc.3",
```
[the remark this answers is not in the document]
mikiw: it's kind of strange but I would leave it as it is, maybe it's a problem in naming, 1.3.3 is confirmed as the last stable version

crates/starknet/src/starknet/mod.rs
taco-paco: Sure

crates/starknet/src/starknet/mod.rs
taco-paco: Sure

crates/starknet-server/tests/test_dump_and_load.rs
```diff
@@ -0,0 +1,272 @@
...
+        }
+    }
+
+    #[tokio::test]
+    async fn mint_dump_on_exit_and_load() {
+        // dump after transaction
+        let dump_file_name = "dump_on_exit";
+        let devnet_dump = BackgroundDevnet::spawn(Some(
+            [
+                "--dump-path".to_string(),
```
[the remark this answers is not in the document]
mikiw: Fixed in 4ab00cfe580744c6f72001bd876ba7d67ff181d9 d78ef69c07c14dc0192dfeafa38820ad9c3ce94e

crates/starknet-server/tests/common/devnet.rs
```diff
@@ -64,7 +64,7 @@ fn get_free_port() -> Result<u16, TestError> {
 impl BackgroundDevnet {
     /// Ensures the background instance spawns at a free port, checks at most `MAX_RETRIES`
     /// times
-    pub(crate) async fn spawn() -> Result<Self, TestError> {
+    pub(crate) async fn spawn(args: Option<Vec<String>>) -> Result<Self, TestError> {
```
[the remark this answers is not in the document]
mikiw: Fixed in 4ab00cfe580744c6f72001bd876ba7d67ff181d9 d78ef69c07c14dc0192dfeafa38820ad9c3ce94e

crates/starknet-server/tests/common/devnet.rs
```diff
@@ -74,6 +74,11 @@ impl BackgroundDevnet {
         let devnet_rpc_url = Url::parse(format!("{}/rpc", devnet_url.as_str()).as_str())?;
         let json_rpc_client = JsonRpcClient::new(HttpTransport::new(devnet_rpc_url.clone()));
 
+        let mut additional_args: Vec<&str> = Vec::new();
+        if let Some(v) = &args {
+            additional_args = v.iter().map(|x| &**x).collect();
+        }
```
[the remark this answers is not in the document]
mikiw: Fixed in 4ab00cfe580744c6f72001bd876ba7d67ff181d9 d78ef69c07c14dc0192dfeafa38820ad9c3ce94e

crates/starknet-server/tests/test_dump_and_load.rs
mikiw: fixed in f3dd993b86e2dfe0e1a3db4d65be92d8f9aa655d

crates/starknet-server/tests/test_dump_and_load.rs
```diff
@@ -0,0 +1,272 @@
...
+        .await
+        .expect("Could not start Devnet");
+        let req_body = Body::from(
+            json!({
+                "address": DUMMY_ADDRESS,
+                "amount": DUMMY_AMOUNT
+            })
+            .to_string(),
+        );
+        let resp = devnet_dump.post_json("/mint".into(), req_body).await.unwrap();
```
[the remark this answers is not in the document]
mikiw: Fixed in b6dead5ef3d8feb2e6407c624126cdc0afd561d3

crates/starknet-server/tests/test_dump_and_load.rs
```diff
@@ -0,0 +1,203 @@
...
+            ["--dump-path", dump_file_name, "--dump-on", "exit"].to_vec(),
+        ))
+        .await
+        .expect("Could not start Devnet");
+        let devnet_dump_pid = devnet_dump.process.id();
+        let mint_tx_hash = devnet_dump.mint(DUMMY_ADDRESS, DUMMY_AMOUNT).await;
+
+        #[cfg(windows)]
+        {
+            // To send SIGINT signal on windows, windows-kill is needed
```
FabijanC, line 79: Ok, but how does a user handle this? They can't press ctrl+c?

mikiw: They can, poblem here was to do it from code level in tests in windows.

crates/starknet-server/tests/test_dump_and_load.rs
mikiw: Fixed in b443161ef05e6b1c2f65fd4e30e7d90460a1abc8 and b443161ef05e6b1c2f65fd4e30e7d90460a1abc8

crates/starknet-server/tests/test_dump_and_load.rs
```diff
@@ -0,0 +1,272 @@
...
+            })
+            .to_string(),
+        );
+        let resp = devnet_dump.post_json("/mint".into(), req_body).await.unwrap();
+        assert_eq!(resp.status(), StatusCode::OK, "Checking status of {resp:?}");
+        let mut resp_body = get_json_body(resp).await;
+        let tx_hash_value = resp_body["tx_hash"].take();
+
+        // Although dump and load is a multiplatform feature and works on all systems is troublesome
+        // on the test level. This test is recommended to run on Linux. There are some problems with
```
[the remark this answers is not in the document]
mikiw: Fixed in b443161ef05e6b1c2f65fd4e30e7d90460a1abc8

crates/starknet/src/starknet/mod.rs
mikiw: I think that this can be implemented without a queue, I'll just attach a new transaction at the end of the file. I also moved the dump logic to another file 42f064dcf59fabf0427e179c8909cc3d6709b204.

crates/starknet/src/starknet/dump.rs
```diff
@@ -0,0 +1,178 @@
...
+                    let mut data = String::new();
+                    file.read_to_string(&mut data).map_err(Error::IoError)?;
+                    let transactions: Vec<Transaction> =
+                        serde_json::from_str(&data).map_err(|_| Error::DeserializationError {
+                            obj_name: "Vec<Transaction>".to_string(),
+                        })?;
+
+                    // to avoid doublets in transaction mode during load, we need to remove the file
+                    // because they will be re-executed and saved again
+                    if self.config.dump_on == Some(DumpMode::OnTransaction) {
```
mikiw, line 166: although it's fixing an issue I'm not sure if I'm satisfied with that, in the other hand detecting that in handle_successful_transaction might be tricky, I'm still checking this.

crates/starknet/src/starknet/dump.rs
```diff
@@ -0,0 +1,178 @@
...
+use starknet_types::rpc::transactions::broadcasted_declare_transaction_v2::BroadcastedDeclareTransactionV2;
+use starknet_types::rpc::transactions::broadcasted_deploy_account_transaction::BroadcastedDeployAccountTransaction;
+use starknet_types::rpc::transactions::broadcasted_invoke_transaction::BroadcastedInvokeTransaction;
+use starknet_types::rpc::transactions::{DeclareTransaction, InvokeTransaction, Transaction};
+
+use super::{DumpMode, Starknet};
+use crate::error::{DevnetResult, Error};
+
+impl Starknet {
+    pub fn re_execute(&mut self, transactions: Vec<Transaction>) -> DevnetResult<()> {
```
mikiw, line 16: All these functions here should be async?

crates/starknet/src/starknet/dump.rs
```diff
@@ -0,0 +1,178 @@
...
+                // can mean that it's first run with dump_path parameter set to dump, in that case
+                // return empty vector
+                if file_path.exists() {
+                    let mut file = File::open(file_path).map_err(Error::IoError)?;
+                    let mut data = String::new();
+                    file.read_to_string(&mut data).map_err(Error::IoError)?;
+                    let transactions: Vec<Transaction> =
+                        serde_json::from_str(&data).map_err(|_| Error::DeserializationError {
+                            obj_name: "Vec<Transaction>".to_string(),
+                        })?;
```
FabijanC, line 162: Should rely on `serde_json::from_reader(file)`

crates/starknet/src/starknet/dump.rs
```diff
@@ -0,0 +1,178 @@
...
+                    let transactions = vec![transaction];
+                    let transactions_dump = serde_json::to_string(&transactions).map_err(|_| {
+                        Error::SerializationError { obj_name: "Vec<Transaction>".to_string() }
+                    })?;
+                    fs::write(Path::new(&path), transactions_dump)?;
+                }
+
+                Ok(())
+            }
+            None => Err(Error::FormatError),
```
FabijanC, line 118: I see that here and in two other places, we just return this FormatError. Do we ever expect it to be thrown? If so, there should be some other info inside to help us debug.

crates/starknet/src/starknet/dump.rs
```diff
@@ -0,0 +1,178 @@
...
+                    let transaction_dump = serde_json::to_string(transaction).map_err(|_| {
+                        Error::SerializationError { obj_name: "Vec<Transaction>".to_string() }
+                    })?;
+                    let mut file = OpenOptions::new()
+                        .append(true)
+                        .write(true)
+                        .open(file_path)
+                        .map_err(Error::IoError)?;
+                    let lenght = file.seek(SeekFrom::End(0)).map_err(Error::IoError)?;
+                    file.set_len(lenght - 1).map_err(Error::IoError)?; // remove last "]" with set_len
```
FabijanC, line 104: Should we somehow assert that the last char is indeed `]`? And log an appropriate message if not.

crates/starknet/src/starknet/dump.rs
FabijanC: I don't understand, which "doublets" does the code comment refer to?

crates/starknet-server/tests/common/devnet.rs
```diff
@@ -64,7 +66,14 @@ fn get_free_port() -> Result<u16, TestError> {
 impl BackgroundDevnet {
     /// Ensures the background instance spawns at a free port, checks at most `MAX_RETRIES`
     /// times
+    #[allow(dead_code)] // dead_code needed to pass clippy
     pub(crate) async fn spawn() -> Result<Self, TestError> {
+        BackgroundDevnet::spawn_with_additional_args(None).await
+    }
+
+    pub(crate) async fn spawn_with_additional_args(
+        args: Option<Vec<&str>>,
```
FabijanC, line 75: This doesn't need to be Option and doesn't need to be a Vec. It would be simpler to just have `&[&str]` (slice of string references).

Then in `spawn()` implementation we would just do `spawn_with_additional_args(&[])`, no need to convert to vectors.

FabijanC (changes requested):

crates/starknet/src/starknet/dump.rs
FabijanC: If CLI argument parsing is done properly, we don't expect this logic branch to ever be reached (the compiler requires it though). Perhaps if #194 will be dealing with this logic, there we can throw a more meaningful error (something that doesn't conceal the real cause and shows both the user and us where the error is coming from).

Marking this as resolved.

crates/starknet/src/starknet/dump.rs
FabijanC: I understand now, upon starting Devnet with a path to load from, re-execution would lead to re-addition of txs into the dump file. It really seems that removing the file the simplest way. Keeping this conversation open to hear other opinions.

crates/starknet-server/tests/common/devnet.rs
mikiw: Fixed in 01f57cd9350b56c38b969e2f005ebf6ee6a8e712

crates/starknet/src/starknet/dump.rs
```diff
@@ -0,0 +1,178 @@
...
+                    // attach to file
+                    let transaction_dump = serde_json::to_string(transaction).map_err(|_| {
+                        Error::SerializationError { obj_name: "Vec<Transaction>".to_string() }
+                    })?;
+                    let mut file = OpenOptions::new()
+                        .append(true)
+                        .write(true)
+                        .open(file_path)
+                        .map_err(Error::IoError)?;
+                    let lenght = file.seek(SeekFrom::End(0)).map_err(Error::IoError)?;
```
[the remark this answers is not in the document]
mikiw: Fixed in efda2284527b3b1ab46a909c8a7c8601bc6225d0

crates/starknet/src/starknet/dump.rs
mikiw: fixed 71e7e3d3d62d9c041f84c577a9d748c15f66a79a

crates/starknet/src/starknet/dump.rs
mikiw: Fixed in b57bd62dcc35e0dc81517883b6bc68d13dc85da8

crates/starknet/src/starknet/dump.rs
```diff
@@ -0,0 +1,187 @@
...
+                let file_path = Path::new(path);
+                if file_path.exists() {
+                    // attach to file
+                    let transaction_dump = serde_json::to_string(transaction).map_err(|_| {
+                        Error::SerializationError { obj_name: "Vec<Transaction>".to_string() }
+                    })?;
+                    let mut file = OpenOptions::new()
+                        .append(true)
+                        .write(true)
+                        .read(true)
```
FabijanC, line 101: I don't remember ever seeing a file being opened in all three modes (append, write, read). Are we sure we need all three of them?

crates/starknet/src/starknet/mod.rs
mikiw: fixed also in 42f064dcf59fabf0427e179c8909cc3d6709b204

crates/starknet/src/starknet/dump.rs
mikiw: I'll check it in https://github.com/0xSpaceShard/starknet-devnet-rs/issues/197

crates/starknet/src/starknet/mod.rs
mikiw: This can be addressed later once you will come back from vacation https://github.com/0xSpaceShard/starknet-devnet-rs/issues/197

crates/starknet/src/starknet/mod.rs
mikiw: This can be addressed later once you will come back from vacation https://github.com/0xSpaceShard/starknet-devnet-rs/issues/197

crates/starknet/src/starknet/dump.rs
mikiw: yes it was needed to add read

mikiw: I was also stunned

FabijanC: but couldn't we just have append and read?

mikiw: write is needed to change size (remove last "]")

FabijanC: I found this: https://stackoverflow.com/a/30684735
It says that since Rust 1.8 `.write(true).append(true)` is the same as `.append(true)`. And really, the docs say the same (we are using Rust 1.69).

mikiw: ok let me check it in code and tests

mikiw: Fixed in 1c5b6e17148162914e6c57101530caab7d288f66

PR #161, opened 2023-09-12
Get storage at return zero instead of NoneState error

marioiordanov:

- get_storage_at returns zero felt, when there is no entry for the provided storage key in the contract
- get_storage_at returns ContractNotFound if the contract address is missing

## Reason for the change
https://github.com/starkware-libs/starknet-specs/blob/6fadca1f1abf48523c0aa799a10061f40f973468/api/starknet_api_openrpc.json#L157

## Checklist:

- [ ] Documented the changes
- [ ] Updated the tests

### DIFF
--- a/crates/starknet/src/account.rs
+++ b/crates/starknet/src/account.rs
@@ -135,15 +135,13 @@ impl Accounted for Account {
 
 #[cfg(test)]
 mod tests {
-    use starknet_in_rust::core::errors::state_errors::StateError;
     use starknet_types::contract_address::ContractAddress;
     use starknet_types::contract_storage_key::ContractStorageKey;
     use starknet_types::felt::Felt;
 
     use super::Account;
-    use crate::error::Error;
     use crate::state::StarknetState;
-    use crate::traits::{Accounted, Deployed};
+    use crate::traits::{Accounted, Deployed, StateChanger};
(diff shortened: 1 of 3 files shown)


crates/starknet/src/account.rs
```diff
@@ -256,17 +254,22 @@ mod tests {
     }
 
     fn setup() -> (Account, StarknetState) {
+        let mut state = StarknetState::default();
+        let fee_token_address = dummy_contract_address();
+
+        state.deploy_contract(fee_token_address, Felt::from(2)).unwrap();
```
FabijanC, line 260: What is `Felt::from(2)`? Class hash? Why can it be any value? I see in another place in this PR it's introduced as `Felt::from(1)` and in another as `dummy_felt()`.

FabijanC (commented):
Implementation looks good, my comments only concern tests.

crates/starknet/src/account.rs
marioiordanov: the class hash in those tests is some arbitrary value. Fixed in [8217105](https://github.com/0xSpaceShard/starknet-devnet-rs/pull/161/commits/8217105b9eb79f9cfeead030994db5b9386ca141) to use the erc20 contract class hash and dummy_felt in the other places

PR #165, opened 2023-09-14
Update sir

marioiordanov:

- Updated starknet_in_rust version to [3febaf6](https://github.com/lambdaclass/starknet_in_rust/commit/3febaf633bfa92caa26d2e1159a503d332b95088).

## Necessary changes

- Modified state_diff implementation, due to this https://github.com/lambdaclass/starknet_in_rust/blob/1eb0411cc875c17c4e176d8411dbf27bb81e3d24/src/transaction/declare_v2.rs#L378
- Changed unit test
- Added unit test

## Checklist:

- [x] Documented the changes
- [x] Updated the tests

### DIFF
--- a/crates/starknet/src/starknet/add_deploy_account_transaction.rs
+++ b/crates/starknet/src/starknet/add_deploy_account_transaction.rs
@@ -140,13 +140,13 @@ mod tests {
             Felt::from(1),
         );
 
-        let sir_transction = transaction
+        let sir_transaction = transaction
             .create_sir_deploy_account(DEVNET_DEFAULT_CHAIN_ID.to_felt().into())
             .unwrap();
 
         // change balance at address
         let account_address =
-            ContractAddress::try_from(sir_transction.contract_address().clone()).unwrap();
+            ContractAddress::try_from(sir_transaction.contract_address().clone()).unwrap();
         let balance_storage_var_address =
             get_storage_var_address("ERC20_balances", &[account_address.into()]).unwrap();
         let balance_storage_key =
--- a/crates/starknet/src/starknet/add_deploy_account_transaction.rs
+++ b/crates/starknet/src/starknet/add_deploy_account_transaction.rs
@@ -104,7 +104,7 @@ mod tests {
     }
 
     #[test]
-    fn deploy_account_transaction_should_fail_due_to_low_balance() {
+    fn deploy_account_transaction_should_fail_due_to_not_enough_balance() {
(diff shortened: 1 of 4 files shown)


crates/starknet/src/starknet/add_deploy_account_transaction.rs
```diff
@@ -121,10 +121,47 @@ mod tests {
...
+            &vec![],
+            Fee(fee_raw),
+            &vec![],
+            Felt::from(0),
+            account_class_hash,
+            Felt::from(13),
+            Felt::from(1),
+        );
+
+        let sir_transction = transaction
```
FabijanC, line 143: typo: transction

crates/starknet/src/starknet/add_deploy_account_transaction.rs
```diff
@@ -121,10 +121,47 @@ mod tests {
         let (txn_hash, _) = starknet.add_deploy_account_transaction(transaction).unwrap();
         let txn = starknet.transactions.get_by_hash_mut(&txn_hash).unwrap();
 
+        assert_eq!(txn.finality_status, None);
+        assert!(txn.execution_result.revert_reason().unwrap().contains("Fee transfer failure"));
+    }
+
+    #[test]
+    fn deploy_account_transaction_should_fail_due_to_not_enough_fee() {
```
FabijanC, line 129: What is the difference between this and the previous test (`deploy_account_transaction_should_fail_due_to_low_balance`)?

FabijanC (changes requested):

crates/starknet/src/starknet/add_deploy_account_transaction.rs
marioiordanov: In this test the account that is making the transaction have some balance in fee_token_contract, but due to the low max fee that it provided, the transaction fails and the max fee is being charged.
In the other one the account doesn't have any balance, so when it tries to charge with the amount = max_fee, it returns an error, something like an exception from starknet_in_rust

crates/starknet/src/starknet/add_deploy_account_transaction.rs
marioiordanov: https://github.com/0xSpaceShard/starknet-devnet-rs/commit/2eb632c3a3230bf568b98ddc0b3f7d974e5d404f

FabijanC (approved):

PR #178, opened 2023-09-21
Complete estimation e2e tests

### LINKED ISSUE #128: Complete estimate fee tests
Currently there are several TODOs in the e2e tests that estimate fee. This can be uncommented and completed once starknet_in_rust fixes estimation on their side.

FabijanC:
## Usage related changes

N/A

- Closes #128
- Fixes the bug in the estimation part of simulate_transactions by introducing `Option<bool>` parameters for passing the flags for charging fee and validating.

## Checklist:

- [x] Checked if in simulation e2e tests we can replace `.ge` with `.gt` (i.e. if SKIP_VALIDATE affects the estimated overall_fee)
- [x] Documented the changes
- [x] Updated the tests


crates/starknet/src/starknet/estimations.rs
```diff
@@ -62,16 +66,24 @@ pub fn estimate_message_fee(
...
+        None,
+        None,
     )
 }
 
 fn estimate_transaction_fee(
     state: &mut StarknetState,
     block_context: &blockifier::block_context::BlockContext,
     transaction: blockifier::transaction::transaction_execution::Transaction,
+    charge_fee: Option<bool>,
```
marioiordanov, line 78: why not be just bool instead of Option.

FabijanC: To have defaulting (unwrap_or). Since both `estimate_fee` and `estimate_message_fee` require constant values (charge_fee=false, validate=true), I would either need to hardcode the constants in both places, or define some global constants and use them in both places (both estimate and estimate_message).

If estimate and estimate_message and really independent, then they deserve to have the constants specified separately, but I at the time of writing, I thought it would be ok to have the same default values for both.

marioiordanov (approved):

PR #202, opened 2023-10-12
Invalid version of DeployAccount transaction

marioiordanov:
Estimating fee of DeployAccount transactions was failing, because blockifier approves only version 1 for DeployAccount transaction. In our implementation the version that is coming from the BroadcastedTransaction is being used. Which for estimate_fee is offset + transaction version. So the fix is to hardcode the version of deploy account transaction to 1, when creating the starknet_api object.

This error is not present in the other types of transaction, because we are not able to assign value to the transaction version.

## Checklist:

- [ ] Documented the changes
- [ ] Updated the tests


ivpavici (approved):

ivpavici:
@marioiordanov @FabijanC I will merge this to see if the tests will pass. 
If there is a different way to solve this, please open another issue for refactor!

Issue #150, opened 2023-08-29
Reverted & rejected txs


taco-paco:
Right now adds rejected txs to transaction list. TODO: implement py devnet's logic with discarding them. Add reverted transactions to block and not leave them in pending state

FabijanC:
Can be used for reference: https://github.com/0xSpaceShard/starknet-devnet/pull/518

FabijanC:
Turns out with the merge of blockifier in #193 the path for this has been largely paved

PR #208, opened 2023-10-13
Fix reverted tx

### LINKED ISSUE #203: State used for storage/nonce retrieval not same as state used for tx execution
There is a problem with handling rejected txs - these transactions should be stored, blocks should be produced and the changed state should be made the new latest state. After merging #193, the state is indeed changed, but block is not produced, and the block is holding the old state with the old nonce - which is causing nonce mismatch.

Related: https://github.com/0xSpaceShard/starknet-devnet-rs/issues/150

## Steps to reproduce
1. Prepare a tx
    - Requires providing nonce: using `starknet_getNonce` to retrieve `account_nonce = N`
3. Estimate the fee of the tx
4. Send a tx with too small fee
    - Expect failure - internally treated as rejected
5. Send a tx with sifficient fee
    - Implicitly using `starknet_getNonce` returns `account_nonce = N`, even though in the changed state it is `N+1`.
    - Error is returned informing of expecting nonce to be `N+1` but `N` was provided.

FabijanC:
## Usage related changes

- #150 still not closeable - we need to bounce back invalid txs, not revert them
- Closes #203

- This will also allow me to continue with #178
- In a few places I renamed `successful` to `accepted` since I believe this is more appropriate (given that we accept even the reverted txs, which aren't successful in the same way as those whose execution has status succeeded).
- In this PR, I can deal with proposed refactoring of commonalities in tx handling.
  - Will not be handled in this PR

## Checklist:

- [x] Documented the changes
- [x] Updated the tests

### DIFF
--- a/crates/starknet/src/starknet/add_declare_transaction.rs
+++ b/crates/starknet/src/starknet/add_declare_transaction.rs
@@ -36,22 +36,16 @@ pub fn add_declare_transaction_v2(
         .execute(&mut starknet.state.state, &starknet.block_context, true, true);
 
     match blockifier_execution_result {
-        Ok(tx_info) => match tx_info.revert_error {
+        Ok(tx_info) => {
             // Add sierra contract
-            Some(error) => {
-                let transaction_to_add =
-                    StarknetTransaction::create_rejected(&transaction, None, &error);
-
-                starknet.transactions.insert(&transaction_hash, transaction_to_add);
-            }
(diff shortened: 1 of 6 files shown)


marioiordanov (commented):
Its straightforward, looks good, what about adding a PR to this PR for refactoring of the results of the transactions.

marioiordanov (approved):

PR #224, opened 2023-10-24
Custom account support

### LINKED ISSUE #183: Feature request: custom account support
Same as the functionality of Pythonic Devnet described here: https://0xspaceshard.github.io/starknet-devnet/docs/guide/accounts#custom-implementation

Old Devnet, however, only supports Cairo 0 accounts, new Devnet could support both Cairo 0 and Cairo 1.

FabijanC:
## Usage related changes

- Introduces new CLI option `--account-class-custom <SIERRA_PATH>`
  - Mutually exclusive with `--account-class [cairo0 | cairo1]`
  - `--account-class-custom` doesn't have a default value, and if no value is provided, we use `--account-class` (which has a default value of `cairo0`)
  - Since Cairo 0 is deprecated, I decided not to have this new CLI option support both custom Cairo 0 and Cairo 1 accounts, but just Cairo 1.
  - There are runtime checks that assert the provided artifact is indeed a Sierra representation of an account (existence of `__execute__` and `__validate__` is asserted).
- Add logging of predeployed accounts class hash to Devnet startup
- Closes #183

- Simplifies `AccountContractClassChoice` by relying on the newly introduced `AccountClassWrapper` (wrapper of class and hash).
  - `AccountContractClassChoice` is needed because it implements `clap::ValueEnum` and allows us to utilize clap's pretty exiting.
- Implements e2e testing by relying on `--account-class cairo1` tests:
  - The existing tests were extracted to functions (with suffix `_body`) which were then used in pairs of tests (one testing `--account-class cairo1`, other testing `--account-class-custom <SIERRA_PATH>`)
  - This is done so because we don't have test parametrization.

## Checklist:

- [x] Documented the changes
- [x] Updated the tests

### DIFF
--- a/crates/starknet-server/src/contract_class_choice.rs
+++ b/crates/starknet-server/src/contract_class_choice.rs
@@ -15,41 +15,27 @@ pub enum AccountContractClassChoice {
     Cairo1,
 }
 
-// TODO refactor to rely on AccountClassWrapper
 impl AccountContractClassChoice {
-    fn get_path(&self) -> &str {
-        match self {
-            AccountContractClassChoice::Cairo0 => CAIRO_0_ACCOUNT_CONTRACT_PATH,
-            AccountContractClassChoice::Cairo1 => CAIRO_1_ACCOUNT_CONTRACT_SIERRA_PATH,
-        }
-    }
-
-    pub fn get_class(&self) -> Result<ContractClass, anyhow::Error> {
-        let contract_class = match self {
-            Self::Cairo0 => ContractClass::Cairo0(Cairo0ContractClass::RawJson(
-                Cairo0Json::raw_json_from_path(self.get_path())?,
-            )),
-            Self::Cairo1 => ContractClass::Cairo1(ContractClass::cairo_1_from_sierra_json_str(
-                std::fs::read_to_string(self.get_path())?.as_str(),
-            )?),
-        };
-        Ok(contract_class)
-    }
-
-    pub fn get_hash(&self) -> Result<Felt, anyhow::Error> {
-        let hash = match self {
+    pub(crate) fn get_class_wrapper(&self) -> Result<AccountClassWrapper, anyhow::Error> {
+        Ok(match self {
             AccountContractClassChoice::Cairo0 => {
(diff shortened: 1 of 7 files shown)


crates/starknet-server/src/contract_class_choice.rs
```diff
@@ -12,39 +16,69 @@ pub enum AccountContractClassChoice {
...
+impl FromStr for AccountClassWrapper {
+    type Err = anyhow::Error;
+
+    fn from_str(path_candidate: &str) -> Result<Self, Self::Err> {
+        // load artifact
+        let contract_class = ContractClass::cairo_1_from_sierra_json_str(
+            std::fs::read_to_string(path_candidate)?.as_str(),
+        )?;
+
+        // check that artifact is really account
```
marioiordanov, line 57: I am not sure if this validation should be here. In my opinion the Starknet::new method have to handle this. The cli validates the format of the arguments

FabijanC: In a way, we are validating user input here (not the input string itself, but what the string refers to).

In case we want to move this validation to the proposed later stage, we need `std::process::exit` logic in `Starknet`, but the only exiting that we've had so far (that I'm aware of) is in CLI argument validation (encapsulated in clap).

marioiordanov: Why there is a need for std::process:exit

FabijanC: We can also exit with `panic`, but I want to avoid that. In any case, we need to somehow exit because Devnet cannot proceed with an invalid class. If it did proceed with a class that e.g. doesn't have `__execute__`, this would be spotted only with the first transaction execution that relies on the predeployed account (too late).

marioiordanov: returning from Starknet::new dont need panicking to stop execution.

marioiordanov: but looking at how the error is formatted and shown to the user, its looking better as It is now

marioiordanov (approved):

PR #234, opened 2023-10-31
feat: l1 l2 messaging postman

### LINKED ISSUE #173: L1-L2 postman functionality should be introduced
As was in devnet-py: https://0xspaceshard.github.io/starknet-devnet/docs/guide/postman

In devnet-py, we depended on a modified version of the Postman subsystem of the cairo-lang Python package, my guess is that something similar will have to exist in our Starknet runtime (starknet_in_rust), and I feel it doesn't.

### LINKED ISSUE #236: messages_sent in receipt is empty
https://github.com/0xSpaceShard/starknet-devnet-rs/blob/72c11f63aa01a6b02a55d09cdcd3cdfe5c58a1cc/crates/types/src/rpc/transactions.rs#L123

glihm:
## Usage related changes

This PR will enable the usage of `postman` endpoints for messaging.

New modules and configurations will be added with this PR. This will affect
how the devnet may be started.

Fixes #173
Fixes #236

## Checklist:

- [X] Applied formatting - `./scripts/format.sh`
- [X] No linter errors - `./scripts/clippy_check.sh`
- [X] Performed code self-review
- [X] Rebased to the last commit of the target branch (or merged it into my branch)
- [X] Documented the changes
- [X] Linked the issues which this PR resolves
- [X] Checked the TODO section in README.md if this PR resolves it
- [X] Updated the tests
- [X] All tests are passing - `cargo test`


crates/starknet/src/transactions.rs
```diff
@@ -211,6 +210,43 @@ impl StarknetTransaction {
             _ => Ok(TransactionReceipt::Common(common_receipt)),
         }
     }
+
+    pub fn get_l2_to_l1_messages(&self) -> Vec<MsgToL1> {
```
FabijanC, line 214: This looks like it could resolve https://github.com/0xSpaceShard/starknet-devnet-rs/issues/236
If so, can you edit your PR description to link it as closeable?

FabijanC: I mean, it could resolve the issue if this line is edited: https://github.com/0xSpaceShard/starknet-devnet-rs/blob/72c11f63aa01a6b02a55d09cdcd3cdfe5c58a1cc/crates/types/src/rpc/transactions.rs#L123

I don't think this conversation should be marked as resolved until that is done (to rely on GitHub's unresolved conversation merge blocker). Let me know if I simply fail to see the responsible commit in the commit history. Also, let's try to supply the commit link to a conversation before closing, e.g. as done here: https://github.com/0xSpaceShard/starknet-devnet-rs/pull/235#discussion_r1377651364

glihm: Oh sorry, totally missed it. You're right I read to fast.
For sure, I'll do.

glihm: Will add testing in the next commits to considered this as a candidate for resolution. :+1:

FabijanC: After merging #233, this PR now has a minor conflict with the base branch. Just replace the TODO comment that I introduced with the code you added to address this conversation.

glihm: Included, thank you @FabijanC for the todo.

crates/starknet/src/transactions.rs
```diff
@@ -159,8 +160,14 @@ impl StarknetTransaction {
...
         let transaction_events = self.get_events();
 
+        let mut transaction_messages: Vec<MessageToL1> = vec![];
+        for message in self.get_l2_to_l1_messages().into_iter() {
+            transaction_messages.push(message.try_into()?);
+        }
+
         let mut common_receipt = self.inner.create_common_receipt(
             &transaction_events,
+            &transaction_messages,
```
glihm, line 170: Due to the argument added by #233, I've now clippy complaining about too many arguments on this function.
What is the strategy you want to adopt? We silent the warning for now? Or do you prefer a rework of the arguments?

crates/types/src/rpc/transactions.rs
```diff
@@ -101,6 +103,7 @@ impl Transaction {
     pub fn create_common_receipt(
         &self,
         transaction_events: &[Event],
+        transaction_messages: &[MessageToL1],
```
glihm, line 106: Due to the argument added by #233, I've now clippy complaining about too many arguments on this function.
What is the strategy you want to adopt? We silent the warning for now? Or do you prefer a rework of the arguments?

crates/starknet-server/src/api/http/models.rs
```diff
@@ -16,18 +18,46 @@ pub(crate) struct LoadPath {
 #[derive(Deserialize, Debug)]
 pub(crate) struct PostmanLoadL1MessagingContract {
     #[serde(rename = "networkUrl")]
-    network_url: String,
-    address: ContractAddress,
+    pub network_url: String,
+    pub address: Option<String>,
+
+    // TODO: THIS IS FOR TESTING PURPOSE ONLY.
```
glihm, line 24: Important point to remove when some configuration can be available from the API.

crates/starknet-server/src/api/http/models.rs
```diff
@@ -16,18 +18,46 @@ pub(crate) struct LoadPath {
...
+    pub l2_contract_address: ContractAddress,
+    #[serde(rename = "entryPointSelector")]
+    pub entry_point_selector: EntryPointSelector,
+    #[serde(rename = "l1ContractAddress")]
+    pub l1_contract_address: ContractAddress,
+    pub payload: Calldata,
+    // TODO: using `u128` as type here will not allow hexadecimal strings. Is that ok?
+    // or do we prefer `Felt` to accept hexadecimal and then convert into `u128`?
+    #[serde(rename = "paidFeeOnL1")]
+    pub paid_fee_on_l1: u128,
```
glihm, line 41: Do we want `u128` here, or we should prefer something that can accept "0x..." values?

crates/types/src/rpc/transactions.rs
FabijanC: If you already know how you would rework the arguments, I'd prefer that. Maybe the whole execution_info could be stored in `fn create_accepted`? Or just the `actual_fee`? Or do you have something else in mind?

crates/starknet-server/src/api/http/models.rs
FabijanC: So this is for the `/postman/send_message_to_l1` endpoint (not governed by the spec)? If so, let's keep it as it was in Devnet-py so far, at least from the user's perspective - probably `Felt` makes the most sense (as your code comment says).

Why would it need to be `u128` internally? If it's really needed, we could use the custom serde deserialization annotation (next to the already used `rename`).

crates/starknet-server/src/api/http/models.rs
FabijanC: Is this a reminder to yourself to remove this before merging? What do you mean by "configuration available from the API"?

glihm: I was referring to this comment found in the code: https://github.com/0xSpaceShard/starknet-devnet-rs/blob/c8af0bbb85b71ad644f4c0f1e040f4b68713d859/crates/starknet-server/src/api/mod.rs#L15.

Does this make sense? Or I'm not in the right and I should read the private key from an already know location?

crates/starknet-server/src/api/http/models.rs
glihm: `u128` it's the type that is used by `starknet_api` under the hood. IMHO, it's important to ensure that we are also bounded to this type to avoid too large values.

This is why in the first try I used `u128` because `axum` will automatically reject the request if it's too large.

But we can totally use `Felt` to accept more values like `hex`, and then we ensure we're not overflowing the `u128` value.

crates/starknet-server/src/api/http/models.rs
FabijanC: the private key of what?

glihm: In order to deploy the messaging contract on L1 node and send transaction to mock the messages on the `MockStarknetMessaging` contract, I do need a private key for the EOA account.

glihm: I mean, if it's only for usage with HardHat, Anvil etc... the private key does not worth that much and can totally be passed in the RPC body along with the other arguments.

crates/starknet-server/src/api/http/models.rs
FabijanC: Two things I care about are:
- keeping the API the same as it was in Devnet-py (meaning: accepting only hex strings for this fee param)
- not making it too complicated for future maintainers (e.g. having as few conversions)

If this can be done using `Felt` - great.

crates/starknet-server/src/api/http/models.rs
FabijanC: Here's what we did in Devnet-py: https://github.com/0xSpaceShard/starknet-devnet/blob/e477aa1bbe2348ba92af2a69c32d2eef2579d863/starknet_devnet/postman_wrapper.py#L158

Is this relevant? IIRC, we were relying on the L1 network to have at least one unlocked account

crates/starknet-server/src/api/http/models.rs
glihm: @FabijanC I introduced a `U128HexOrDec` visitor for `serde`. I've tried a first implementation with `Felt`, but it not that obvious for the reader of the models to know that under the hood it's limited to `u128` values range.

With this `serde` visitor, we ensure that we can accept decimal or hexadecimal string, without compromising the readability of the model and the range check on the 

The related commit is this one: 90a70097659b478af5d86b22408391015141696a

please let me know if it makes sense to your too. If not, I'll be happy to revisit this.

crates/starknet-server/src/api/http/models.rs
glihm: From the python documentation here https://web3py.readthedocs.io/en/stable/web3.eth.account.html#local-vs-hosted-keys it looks like they are pre-defining the keys for the corresponding node in the code.

In the case of `ethers`, I think they are agnostic of the local node used for testing.

So we can totally replicate what `web3` in python is doing, by pre-registering in the messaging struct some accounts, for `HardHat` and `Anvil` at least.
But this means, we've to update the code anytime we want to support an other tool.

1. It's not a problem and we start by having this list somewhere in the code.
2. We can do both, to ensure a good evolution of Devnet without the need to recompile. If the user provides a `private_key` in the `postman_load` endpoint, it's fine we use it. If no `private_key` is given (like in devnet-py), we attempt to deploy with the first default account of `HardHat`/`Anvil`/...

WDYT?

crates/starknet-server/src/api/http/models.rs
FabijanC: Since now decimal strings are also accepted, isn't this an expansion of the API? IMO having just one supported format is less error-prone.

Also, what is the advantage of introducing the new dependency (serde_with) when we can achieve this with existing dependencies? E.g. we already use `serialize_with`, `deserialize_with` and `with` in serde struct annotations.

crates/starknet-server/src/api/http/models.rs
FabijanC: I like the suggestion in 2. We might also need to pass the account address together with the private key?

glihm: Yes sure the address is always passed like the devnet py API. We just have this extension with the private key that is optional.
If the address passed is known as pre-registered, we use a private key we have internally.
If the address is not known we expected a private key along the network url. We can then send bad request or more appropriate status.

FabijanC: Sounds good

crates/starknet-server/src/api/http/models.rs
glihm: >Since now decimal strings are also accepted, isn't this an expansion of the API? IMO having just one supported format is less error-prone.
In that case, if you prefer to be 100% compatible with the devnet-py API, I'll use `Felt` as any other type. This is less correct for the model, but better for simplicity.

If it's not a problem to change a bit from devnet-py, we can use raw `u128` as `MintTokenRequest` is accepting.

>Also, what is the advantage of introducing the new dependency (serde_with) when we can achieve this with existing dependencies? E.g. we already use serialize_with, deserialize_with and with in serde struct annotations.

My bad on this, the habit of doing it with `serde_with` and the `Visitor`. Will remove it.

crates/types/src/rpc/transactions.rs
glihm: Not sure yet about this one, trying to get the big picture of possible rework.
The first thing I see is that as we already import some `rpc` stuff into `starknet` crate, we may choose to directly pass the `TransactionOutput` instead of passing `events`, `messages` and `actual_fee` separately. This will reduce the number of args by 2.

But then, I was trying to see if it's valuable to relocate the `create_common_receipt` as lot's of arguments are coming the `starknet` transaction type. And `starknet` can import `rpc`, but not the opposite.

WDYT?

crates/starknet-server/src/api/http/models.rs
glihm: > If this can be done using Felt - great.

To follow this, the commit 236ce995f707a06dbdc25316d35574726300a790 is working with `Felt` only. Let me know if it's ok that way.

crates/types/src/rpc/transactions.rs
glihm: I attempted in this commit b23e9e3e42910f99eb296d979a7320d8198f66ca to pass the transaction output instead of each individual elements.

If it's fine for now, good. If you prefer a rework by moving the `create_common_receipt`, or any other, let me know. :+1:

crates/starknet-server/src/api/http/models.rs
glihm: After testing a bit, this commit 9dfe0c980457bb20179d8ede8bd2552666191dc2 introduces the behavior very similar to devnet-py.
If no `private_key` passed in the model of `postman/load_l1_messaging_contract`, we use the `accounts[0]` of the default wallet for eth devnets (at least hardhat and anvil compatible).

And to avoid problem with new eth devnet or other configuration of eth devnet the user may have, we accept (as an extension from devnet-py) a `private_key` (which is optional).
If this `private_key` is given, it is used.

We don't need the account address as the only computation we need is signing transaction, where only the private key is required.

Good for you @FabijanC?

crates/types/src/rpc/transactions.rs
FabijanC: Fine for now, resolving the conversation

crates/starknet-server/src/api/http/models.rs
FabijanC: Looks ok, resolving

crates/starknet-server/src/api/http/models.rs
FabijanC: 1. Any reason why you are not defaulting to `ETH_ACCOUNT_DEFAULT.private_key` using `.unwrap_or`?
2. But more importantly, I don't see why we now need this dummy `ETH_ACCOUNT_DEFAULT` when it wasn't needed in devnet-py; there we just relied on `web3.eth.accounts` (likely relying on the [eth_accounts RPC method](https://docs.infura.io/networks/ethereum/json-rpc-methods/eth_accounts))  having the account at index 0 unlocked.

mikiw:
All solidity contracts are taken from https://github.com/starknet-io/starkgate-contracts/tree/main/src/starkware/starknet/solidity? Can we also document in readme.md how json artifact/artifacts are generated for future developers? Sooner or later this can change and we will need to make an update.

crates/starknet/contracts/artifacts/MockStarknetMessaging.json
```diff
@@ -0,0 +1,2599 @@
+{
```
mikiw, line 1: Do we need this json here? we already have this artifact in contracts folder

glihm: > All solidity contracts are taken from https://github.com/starknet-io/starkgate-contracts/tree/main/src/starkware/starknet/solidity?

Nope, for now I took exactly what you had in the devnet-py. As I didn't know if I was allowed to modify them.

> Can we also document in readme.md how json artifact/artifacts are generated for future developers? Sooner or later this can change and we will need to make an update.

Yes this is planned, everything will be documented once the definitive location/organization of files is approved.

> Do we need this json here? we already have this artifact in contracts folder

This is one comment I made in the code, due to how cargo works, this location of the contracts is relative to the `Cargo.toml` file. So when you execute at workspace level / package level, the resolution is different. I left a comment as this needs to be solved. I am perhaps not using what I need to use, please feel free to point out what I'm surely missing out then. :+1:

crates/starknet-server/src/api/http/models.rs
glihm: Oh.. I didn't know about this RPC method.
But this method is only returning the address of the account for what I've just tested, not it's private key. This leads to me second point below.

About unlocking account I don't know (yet) how to do it I've never used it, and the `ethers` methods are asking for a signer. I'll then investigate how this can be done.

If you have any knowledge on this, I'll be happy to learn about it.

glihm: > Any reason why you are not defaulting to ETH_ACCOUNT_DEFAULT.private_key using .unwrap_or?

nope, good catch. I had more code here before, didn't refacto. Thanks for the remark.

crates/starknet/contracts/artifacts/MockStarknetMessaging.json
glihm: Path resolution solved in d8bfca76e6402e872f2552a18ef246bae1adf2e2

glihm: Documentation solved in 0eea51d444ad586125647bfa27a722cbec5a43fc

contracts/README.md
```diff
@@ -0,0 +1,180 @@
+# Devnet contracts for development
+
+This folder contains Cairo and Solidity contracts for developement of Devnet.
+If you wish you to check specifically one of the two chains contracts, please check:
+1. `solidity` folder for Ethereum related contracts.
+2. `cairo` folder for Starknet related contracts, and example of how to work with starknet without
+   running a L1 node.
+
```
ivpavici, line 8: note for devx:
we should note here somewhere the existence of readme.md files in both /cairo and /solidity , because there are instructions on installing Scarb etc.
Also to list all dependencies (like starkli), links to them etc.

contracts/README.md
```diff
@@ -0,0 +1,180 @@
...
+{
+    "messageContractAddress":"0x5fbdb2315678afecb367f032d93f642f64180aa3"
+}
+```
+
+2. Deploy the `L1L2.sol` contract in order to receive/send messages from/to L2.
+```bash
+# Ethereum terminal (contracts/solidity)
+source .env
+forge script script/L1L2.s.sol:Deploy --broadcast --rpc-url $ETH_RPC_URL --silent
```
ivpavici, line 56:
```suggestion
forge script script/L1L2.s.sol:Deploy --broadcast --rpc-url $ETH_RPC_URL 
```
...to make the below output visible

contracts/README.md
```diff
@@ -0,0 +1,180 @@
...
+```bash
+anvil
+```
+
+For Starknet, ensure you've Devnet compiled and running with the following params:
+```bash
+cargo run -- --seed 42
+```
+
+Now both nodes are running, Devnet for Starkne and Anvil for Ethereum.
```
ivpavici, line 22:
```suggestion
Now both nodes are running, Devnet for Starknet and Anvil for Ethereum.
```

contracts/README.md
```diff
@@ -0,0 +1,180 @@
...
+```
+
+### Starknet contracts and send message to L1
+1. On the Starknet terminal, set [starkli](https://book.starkli.rs/installation) variables:
+```bash
+# Starknet terminal (contracts/cairo)
+source .env
+
+# Declare
+starkli declare target/dev/cairo_l1_l2.contract_class.json
```
ivpavici, line 80: we can probably pre-deploy this contract on the devnet startup, thus saving these steps!

contracts/solidity/.github/workflows/test.yml
```diff
@@ -0,0 +1,34 @@
+name: test
```
ivpavici, line 1: mentioned this on discord but just not to forget, this file can be removed

contracts/README.md
```diff
@@ -0,0 +1,180 @@
+# Devnet contracts for development
+
```
ivpavici, line 2: I would suggest also to rename the folder from /contracts to maybe /l1-l2-e2e or something better

contracts/generate_artifacts.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
```
ivpavici, line 1: is it maybe more intuitive to put this script inside the /solidity folder ?

crates/starknet/Cargo.toml
```diff
@@ -11,6 +11,11 @@ blockifier = { workspace = true, features = ["testing"]}
 cairo-felt = { workspace = true }
 cairo-lang-starknet = { workspace = true }
 clap = { workspace = true }
+
+# TODO: perhaps not the best solution to have ethers inside Starknet crate.
+# Here for the MVP, to be checked with spaceshard team.
+ethers = { version = "2.0.8" }
```
mikiw, line 17: for me, it's ok since it's needed for l1l2, I would remove this todo, @FabijanC please also confirm that

mikiw: in old devnet we use web3 library

FabijanC: Maybe @glihm was worried about the location of this dependency (i.e. it not being defined in the root Cargo.toml). If that is the concern, I summon @marioiordanov

crates/starknet/src/starknet/add_l1_handler_transaction.rs
```diff
@@ -0,0 +1,262 @@
...
+                    selector,
+                )),
+            )) => {
+                assert_eq!(selector.0, withdraw_selector.into())
+            }
+            _ => panic!("Wrong error type"),
+        }
+    }
+
+    // TODO: call withdraw and verify that a message is present in the state.
```
mikiw, line 142: I think that this can be removed

crates/starknet/src/starknet/mod.rs
```diff
@@ -69,10 +71,16 @@ pub struct Starknet {
     pub(in crate::starknet) state: StarknetState,
     predeployed_accounts: PredeployedAccounts,
     pub(in crate::starknet) block_context: BlockContext,
-    blocks: StarknetBlocks,
+    // TODO: to avoid repeating some logic from blocks,
```
mikiw, line 74: I would remove todo but I would keep the comment.

crates/starknet/src/transactions.rs
```diff
@@ -157,15 +158,26 @@ impl StarknetTransaction {
...
+
+        let mut messages_sent: Vec<MessageToL1> = vec![];
+        for message in self.get_l2_to_l1_messages().into_iter() {
+            messages_sent.push(message.try_into()?);
+        }
+
+        let output = TransactionOutput {
+            actual_fee: self.execution_info.actual_fee,
+            messages_sent,
+            // TODO: we may have performances issues here? Can have a lot's of events..
```
mikiw, line 171: I would keep the comment but I would remove TODO. If in the future there will be a performance issue related to l1l2, we will take a look here anyway.

crates/types/src/rpc/transactions/l1_handler_transaction.rs
```diff
@@ -0,0 +1,128 @@
...
+const PREFIX_L1_HANDLER: FieldElement = FieldElement::from_mont([
+    1365666230910873368,
+    18446744073708665300,
+    18446744073709551615,
+    157895833347907735,
+]);
+
+impl L1HandlerTransaction {
+    /// Instanciates a new `L1HandlerTransaction`.
+    /// TODO: a bit too much arguments here...
```
mikiw, line 30: I think that this TODO can be removed

crates/types/src/rpc/transactions/l1_handler_transaction.rs
```diff
@@ -0,0 +1,128 @@
...
+                    .calldata
+                    .iter()
+                    .map(|felt| FieldElement::from(*felt))
+                    .collect::<Vec<FieldElement>>(),
+            ),
+            fee,
+            // TODO: How to get the chain_id at this point?
+            // Or should this be computed in an other fashion?
+            // chain_id.into(),
+            0_u128.into(),
```
mikiw, line 123: This might be problematic... I'll take a closer look tomorrow.

mikiw: Can we just add chain_id in L1HandlerTransaction and mark it as serde skip?
```
    #[serde(skip)]
    pub chain_id: u128,
```

@marioiordanov @FabijanC

mikiw (changes requested):
There is one TODO that I think should be handled in this PR which is hardcoded chain_id.

I also added comments to TODO's that in my opinion can be removed.

Documentation in readme files are good. I tested locally l1l2 scenario on my mac and it's working fine :thumbsup:

good job @glihm ! :fire::fire::fire:

crates/starknet-server/src/api/http/endpoints/postman.rs
```diff
@@ -1,30 +1,158 @@
...
+            .await
+            .map_err(|e| HttpApiError::MessagingError { msg: e.to_string() })?
+            .into_iter()
+            .map(|m| m.try_into())
+            .collect::<Result<Vec<MessageToL2>, HttpApiError>>()
+    };
+
+    let from_block = if let Some(m) = &starknet.messaging { m.last_local_block } else { 0 };
+
+    let (messages_to_l1, last_local_block) = if dry_run {
```
FabijanC, line 54: Under this line, there are some cryptic variable names. I'm ok with `e` for error, but `m` and `b` are not intuitive.

crates/starknet-server/src/api/http/endpoints/postman.rs
```diff
@@ -1,30 +1,158 @@
...
+                    l1_provider,
+                }))
+            } else {
+                Err(HttpApiError::MessagingError { msg: format!("MessagesToL1: {}", e) })
+            }
+        }
+        (Err(e1), Err(e2)) => Err(HttpApiError::MessagingError {
+            msg: format!("MessagesToL1: {} || MessagesToL2: {}", e1, e2),
+        }),
+    }
```
FabijanC, line 120: I notice flush is now returning Error if it cannot flush, whereas devnet-py simply returned `{"generated_l2_transactions":[]}`. In fact, the whole response of `/postman/flush` seems to be changed. Do you think devnet-py had it wrong?

crates/starknet-server/src/api/http/endpoints/postman.rs
```diff
@@ -1,30 +1,158 @@
...
+    };
+
+    if !dry_run {
+        // +1 to ensure this last block is not collected anymore.
+        starknet.messaging.as_mut().expect("Messaging expected configured").last_local_block =
+            last_local_block + 1;
+    }
+
+    match (messages_to_l1, messages_to_l2) {
+        (Ok(l1s), Ok(l2s)) => {
```
FabijanC, line 92: Why does this need to be four cases? Can't we first process `messages_to_l1`, then `messages_to_l2`?

FabijanC (changes requested):

crates/starknet/src/transactions.rs
marioiordanov: @glijm feel free to remove the `clone` method invocation

crates/starknet-server/tests/test_messaging.rs
```diff
@@ -0,0 +1,229 @@
...
+    use starknet_rs_core::types::{BlockId, BlockTag, FieldElement, FunctionCall};
+    use starknet_rs_core::utils::{
+        get_selector_from_name, get_udc_deployed_address, UdcUniqueness,
+    };
+    use starknet_rs_providers::jsonrpc::HttpTransport;
+    use starknet_rs_providers::{JsonRpcClient, Provider};
+    use starknet_rs_signers::LocalWallet;
+
+    use crate::common::constants::{CHAIN_ID, MESSAGING_L1_ALLOWED_CONTRACT};
+    use crate::common::devnet::BackgroundDevnet;
```
FabijanC, line 33: Note that `crate::common::devnet` has been renamed to `crate::common::background_devnet`

crates/starknet-server/src/api/http/models.rs
glihm: @FabijanC to follow up on this, do you want to remove `private_key` and we give to the user the expected accounts? (which are already setup by default for anvil and hardhat).

contracts/README.md
glihm: Good point, the updated README is here: https://github.com/glihm/starknet-devnet-rs/tree/7dd5d6e0fe8f14aca40ebd562cc3c5d5e16ce873/contracts/l1-l2-messaging with now an explicit reference to other READMEs.

contracts/README.md
glihm: Updated: https://github.com/glihm/starknet-devnet-rs/tree/7dd5d6e0fe8f14aca40ebd562cc3c5d5e16ce873/contracts/l1-l2-messaging

contracts/README.md
glihm: Fixed in the updated [README](https://github.com/glihm/starknet-devnet-rs/tree/7dd5d6e0fe8f14aca40ebd562cc3c5d5e16ce873/contracts/l1-l2-messaging)

contracts/README.md
glihm: I would also tend to follow @FabijanC point of view. The flow already includes the generation of the file.

contracts/solidity/.github/workflows/test.yml
glihm: Removed.

contracts/README.md
glihm: Now [contracts folder](https://github.com/glihm/starknet-devnet-rs/tree/7dd5d6e0fe8f14aca40ebd562cc3c5d5e16ce873/contracts) contains sub-folders with important files, and `l1-l2-messaging` sub-folder for messaging related contracts. Is that ok?

contracts/generate_artifacts.sh
glihm: @FabijanC exactly, even if they are not used directly for compiling (at least for now), you can now find the associated actions [in the script](https://github.com/glihm/starknet-devnet-rs/blob/7dd5d6e0fe8f14aca40ebd562cc3c5d5e16ce873/contracts/generate_artifacts.sh).
Also, running this script is now **mandatory** before building Devnet. The readme was adjusted accordingly.

crates/starknet/Cargo.toml
glihm: @FabijanC correct. But it looks like ok for now, let me know what do you think @marioiordanov.

crates/starknet/src/starknet/mod.rs
glihm: Fixed in e994a0fe9e906b077dcea7df34f9a74c854e815c.

crates/types/src/rpc/transactions/l1_handler_transaction.rs
glihm: Awaiting the final answer, this implementation has been removed. Please don't hesitate after a decision is made if I need to put it back again. :+1:

rust-toolchain.toml
```diff
@@ -1,5 +1,5 @@
 [toolchain]
-channel = "1.69.0"
+channel = "1.70.0"
```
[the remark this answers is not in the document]
glihm: This is the error I have with `1.69.0`, which is gone using `1.70.0`.

```
error: package `clap_builder v4.4.8` cannot be built because it requires rustc 1.70.0 or newer, while the currently active rustc version is 1.69.0
Either upgrade to rustc 1.70.0 or newer, or use
cargo update -p clap_builder@4.4.8 --precise ver
where `ver` is the latest version of `clap_builder` supporting rustc 1.69.0
```

contracts/README.md
```diff
@@ -0,0 +1,180 @@
...
+}
+```
+
+2. Deploy the `L1L2.sol` contract in order to receive/send messages from/to L2.
+```bash
+# Ethereum terminal (contracts/solidity)
+source .env
+forge script script/L1L2.s.sol:Deploy --broadcast --rpc-url $ETH_RPC_URL --silent
+```
+```bash
```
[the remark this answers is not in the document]
glihm: Removed in latest [README](https://github.com/glihm/starknet-devnet-rs/blob/92e2d9d3ec22bd1db06e575ff687a672dfa60546/contracts/l1-l2-messaging/README.md).

contracts/README.md
```diff
@@ -0,0 +1,180 @@
...
+            "payload":["0x0","0x1","0x1"]
+        }
+    ],
+    "messagesToL2":[],
+    "l1Provider":"dry run"
+}
+```
+```bash
+# Flushing the message to actually send them to the L1.
+curl -H 'Content-Type: application/json' -d '{}' http://127.0.0.1:5050/postman/flush
```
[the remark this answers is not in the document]
glihm: Fixed in 0b480ed50c0e7a5d98b7b8ec7a2d04567051ac23 to have an optional body. This yields the following curl command:
```
curl -H 'Content-Type: application/json' -X POST http://127.0.0.1:5050/postman/flush
```

contracts/solidity/.env.anvil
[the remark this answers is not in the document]
glihm: I refactored by having only one `.env`, please let me know if it's clearer executing README operations now.
IMHO, it is good to keep an additional `.env` in each folder in order to provider an example for each side. If you prefer to remove them and only keep one, let me know and I'll remove those in `cairo` and `solidity` folders.

contracts/solidity/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+First, install forge following the [official book](https://book.getfoundry.sh/getting-started/installation).
+
+### Build
+
+To build the contracts, enter the solidity directory and run:
+```bash
+forge build
+```
+This will generates the artifacts of the contracts into the `out` folder.
```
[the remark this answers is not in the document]
glihm: Removed.

crates/types/src/rpc/transaction_receipt.rs
```diff
@@ -82,6 +84,25 @@ pub struct MessageToL1 {
     pub payload: L2ToL1Payload,
 }
 
+impl TryFrom<MsgToL1> for MessageToL1 {
```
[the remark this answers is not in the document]
glihm: I refactored all the types related to messaging into [crates/types/src/rpc/messaging.rs](https://github.com/glihm/starknet-devnet-rs/blob/7dd5d6e0fe8f14aca40ebd562cc3c5d5e16ce873/crates/types/src/rpc/messaging.rs) to simplify the management of those types. This also removed the duplication of types, and we use those types the most we can to facilitate code maintenance.

For symmetry, `MessageToL1` and `MessageToL2` looks more adapted.

Sounds good to you?

crates/starknet-server/src/api/http/models.rs
```diff
@@ -16,30 +19,117 @@ pub(crate) struct LoadPath {
 #[derive(Deserialize, Debug)]
 pub(crate) struct PostmanLoadL1MessagingContract {
     #[serde(rename = "networkUrl")]
```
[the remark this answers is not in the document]
glihm: Love that. I removed all the `serde` renaming to use `snake_case` for all postman endpoints. You can find the new types in [messaging.rs](https://github.com/glihm/starknet-devnet-rs/blob/7dd5d6e0fe8f14aca40ebd562cc3c5d5e16ce873/crates/types/src/rpc/messaging.rs) that are directly used by postman endpoints.

crates/types/src/rpc/transactions/l1_handler_transaction.rs
```diff
@@ -0,0 +1,128 @@
...
+        };
+
+        Ok(transaction)
+    }
+}
+
+// TODO: for this version I didn't impl `HashProducer`
+// as the chain_id is not present in the transaction itself.
+// Should we keep the chain_id somewhere in the devnet `L1HandlerTransaction`?
+// Or we don't want HashProducder on `L1HandlerTransaction` and the `new` is fine?
```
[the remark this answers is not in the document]
glihm: I removed it, related to the @mikiw comment. I'll wait a decision on that before adding it again if required. :+1:

crates/starknet-server/tests/test_messaging.rs
```diff
@@ -0,0 +1,229 @@
...
+        assert_eq!(payload[1], "0x1");
+        // Amount.
+        assert_eq!(payload[2], "0x1");
+
+        assert_eq!(body.get("messagesToL2").unwrap().as_array().unwrap().len(), 0);
+        assert_eq!(body.get("l1Provider").unwrap().as_str().unwrap(), "dry run");
+    }
+
+    #[tokio::test]
+    async fn can_receive_mock_message_to_l2() {
```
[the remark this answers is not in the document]
glihm: Yes, good point. I'll need to take time on this one. We should be able to call it via Rust, but I need to confirm that.

crates/types/src/rpc/transactions/l1_handler_transaction.rs
```diff
@@ -0,0 +1,128 @@
...
+    pub fn create_blockifier_transaction(&self) -> DevnetResult<BlockifierL1HandlerTransaction> {
+        let version: Felt = 0_u128.into();
+
+        let transaction = BlockifierL1HandlerTransaction {
+            tx: ApiL1HandlerTransaction {
+                contract_address: ApiContractAddress::try_from(self.contract_address)?,
+                entry_point_selector: ApiEntryPointSelector(self.entry_point_selector.into()),
+                calldata: ApiCalldata(Arc::new(self.calldata.iter().map(|f| f.into()).collect())),
+                nonce: ApiNonce(self.nonce.into()),
+                version: ApiTransactionVersion(version.into()),
```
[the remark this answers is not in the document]
glihm: I've updated the code [here](https://github.com/glihm/starknet-devnet-rs/blob/7dd5d6e0fe8f14aca40ebd562cc3c5d5e16ce873/crates/types/src/rpc/transactions/l1_handler_transaction.rs#L41) and added a comment.

It's mostly because only version 0 is supported for now, and I wanted to ensure that this is limited to the value `0` until upgraded voluntarily by the developer.

If it's not good for you, I can definitely remove this assert. IMHO it's better to enforce the check as this is the only value that must be supported for now.

crates/starknet/src/transactions.rs
```diff
@@ -194,6 +206,41 @@ impl StarknetTransaction {
...
+            messages.extend(get_blockifier_messages_recursively(info));
+        }
+
+        if let Some(info) = self.execution_info.execute_call_info.as_ref() {
+            messages.extend(get_blockifier_messages_recursively(info));
+        }
+
+        if let Some(info) = self.execution_info.fee_transfer_call_info.as_ref() {
+            messages.extend(get_blockifier_messages_recursively(info));
+        }
```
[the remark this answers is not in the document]
glihm: Refactored [during the merge](https://github.com/glihm/starknet-devnet-rs/blob/7dd5d6e0fe8f14aca40ebd562cc3c5d5e16ce873/crates/starknet/src/transactions.rs#L221) with main to follow the way it's done for events.

crates/starknet/src/starknet/add_l1_handler_transaction.rs
```diff
@@ -0,0 +1,262 @@
...
+pub fn add_l1_handler_transaction(
+    starknet: &mut Starknet,
+    transaction: L1HandlerTransaction,
+) -> DevnetResult<TransactionHash> {
+    let transaction_hash = transaction.transaction_hash;
+    trace!("Executing L1 handler transaction [{:#064x}]", transaction.transaction_hash);
+
+    let blockifier_transaction = transaction.create_blockifier_transaction()?;
+
+    let charge_fee = false;
```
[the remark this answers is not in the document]
glihm: The fees are charged on L1 as the `L1HandlerTransaction` is not executed by an account. Nobody can be charged. We can set it to true, the outcome should be the same.
I added a comment [here in the code](https://github.com/glihm/starknet-devnet-rs/blob/7dd5d6e0fe8f14aca40ebd562cc3c5d5e16ce873/crates/starknet/src/starknet/add_l1_handler_transaction.rs#L18) to clarify it.

crates/starknet/src/messaging/ethereum.rs
```diff
@@ -0,0 +1,536 @@
...
+mod tests {
+    use starknet_types::chain_id::ChainId;
+
+    use super::*;
+
+    #[test]
+    fn l1_handler_tx_from_log_parse_ok() {
+        let from_address = "0x000000000000000000000000be3C44c09bc1a3566F3e1CA12e5AbA0fA4Ca72Be";
+        let to_address = "0x039dc79e64f4bb3289240f88e0bae7d21735bef0d1a51b2bf3c4730cb16983e1";
+        let selector = "0x02f15cff7b0eed8b9beb162696cf4e3e0e35fa7032af69cd1b7d2ac67a13f40f";
```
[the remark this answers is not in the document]
glihm: Good point, I added a comment [in the code](https://github.com/glihm/starknet-devnet-rs/blob/7dd5d6e0fe8f14aca40ebd562cc3c5d5e16ce873/crates/starknet/src/messaging/ethereum.rs#L468). They are hard-coded values coming from a transaction on Goerli.

crates/starknet/src/messaging/mod.rs
```diff
@@ -0,0 +1,189 @@
...
+    /// for each one of them.
+    pub async fn fetch_and_execute_messages_to_l2(
+        &mut self,
+    ) -> DevnetResult<Vec<L1HandlerTransaction>> {
+        if self.messaging.is_none() {
+            return Err(Error::MessagingError(MessagingError::NotConfigured));
+        }
+
+        let chain_id = self.chain_id().to_felt();
+        let messaging = self.messaging.as_mut().unwrap();
```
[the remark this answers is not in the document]
glihm: Good point, [I added a getter](https://github.com/glihm/starknet-devnet-rs/blob/7dd5d6e0fe8f14aca40ebd562cc3c5d5e16ce873/crates/starknet/src/messaging/mod.rs#L62) to ease the retrieval of this value, with a `Result` instead of an `expect`.

crates/starknet/src/messaging/ethereum.rs
```diff
@@ -0,0 +1,536 @@
...
+    pub async fn fetch_messages(
+        &mut self,
+        chain_id: Felt,
+    ) -> DevnetResult<Vec<L1HandlerTransaction>> {
+        let chain_latest_block: u64 = self
+            .provider
+            .get_block_number()
+            .await?
+            .try_into()
+            .expect("Can't convert latest block number into u64.");
```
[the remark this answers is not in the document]
glihm: Fixed in e994a0fe9e906b077dcea7df34f9a74c854e815c with result instead of expect.

crates/starknet/src/messaging/ethereum.rs
```diff
@@ -0,0 +1,536 @@
...
+            .get_logs(&filters)
+            .await?
+            .into_iter()
+            .filter(|log| log.block_number.is_some())
+            .map(|log| {
+                (
+                    log.block_number
+                        .unwrap()
+                        .try_into()
+                        .expect("Block number couldn't be converted to u64."),
```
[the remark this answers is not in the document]
glihm: Fixed in e994a0fe9e906b077dcea7df34f9a74c854e815c with Result instead of expect.

crates/starknet-server/src/api/http/endpoints/postman.rs
glihm: I actually took the output from the video you sent to me on Discord. And there were no `generated_l2_transactions`. But it's a very good point so I added it into the flush result.

```rust
#[derive(Serialize)]
pub(crate) struct FlushedMessages {
    pub messages_to_l1: Vec<MessageToL1>,
    pub messages_to_l2: Vec<MessageToL2>,
    pub generated_l2_transactions: Vec<TransactionHash>,
    pub l1_provider: String,
}
```
I do think that only having the transaction hashes is too low for a devX perspective when debugging messaging. But having those hashes makes a lot's of sense.

Also, now if there is an error, I don't think we should return empty arrays. Having the corresponding error is required as the user will know exactly where the error happened and he will be able to solve it (or for us very helpful to debug ^^).

crates/starknet-server/src/api/http/endpoints/postman.rs
glihm: I removed this it was not necessary as you mentioned.

crates/starknet-server/tests/test_messaging.rs
glihm: Thanks for the info, was adjusted accordingly.

contracts/solidity/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+### Testing
```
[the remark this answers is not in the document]
glihm: Fixed.

crates/starknet/src/messaging/ethereum.rs
```diff
@@ -0,0 +1,536 @@
...
+            provider_signer: Arc::new(provider_signer),
+            messaging_contract_address: Address::zero(),
+            last_fetched_block: 0,
+            last_local_block: 0,
+        };
+
+        if let Some(address) = contract_address {
+            ethereum.messaging_contract_address = Address::from_str(address).map_err(|e| {
+                Error::MessagingError(MessagingError::EthersError(format!(
+                    "Address can't be parse from string: {} ({})",
```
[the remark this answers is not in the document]
glihm: Fixed in c77425dfb91d85f7b548ba5c62a41ea350e4e10e.

crates/starknet-server/tests/test_messaging.rs
```diff
@@ -0,0 +1,229 @@
+//! This module tests the messaging feature from the starknet-server perspective.
+//!
+//! The message contract `cairo_0_l1l2_contract` associated a balance to a user (ContractAddress)
```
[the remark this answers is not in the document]
glihm: Fixed in c77425dfb91d85f7b548ba5c62a41ea350e4e10e.

crates/types/src/rpc/transactions/l1_handler_transaction.rs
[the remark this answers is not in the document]
FabijanC: So far we've avoided using `assert` in production code. We've relied on using `Err`, i.e. `Result`.

If we will have some kind of check, then why not use `self.version` but define a new variable in `create_blockifier_transaction`. That way we have to remember to apply the future change in two places (this method and `compute_hash`).

Also, why should the check be done only on hash computation (i.e. in `compute_hash`)?

crates/starknet/src/starknet/add_l1_handler_transaction.rs
```diff
@@ -0,0 +1,262 @@
...
+        match result.err().unwrap() {
+            crate::error::Error::BlockifierTransactionError(ExecutionError(
+                EntryPointExecutionError::PreExecutionError(PreExecutionError::EntryPointNotFound(
+                    selector,
+                )),
+            )) => {
+                assert_eq!(selector.0, withdraw_selector.into())
+            }
+            _ => panic!("Wrong error type"),
+        }
```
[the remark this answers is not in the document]
FabijanC: There's redundancy, no need to use `assert!(result.is_err())` since it's covered by `match`. If this `is_err` assertion fails, we don't see the exact result.

crates/starknet/src/transactions.rs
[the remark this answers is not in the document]
FabijanC: This doesn't address my observation about unnecessarily manually enumerating all of the execution info types, it just adds more code duplication.

crates/starknet-server/src/api/http/endpoints/postman.rs
FabijanC: Sure if there is an error we shold inform the user of the error. But my real question is, is it really an error if the users wishes to flush when there is nothing to flush? Seems like we used to treat it as an idempotent operation. That's what I asked if you consider wrong.

crates/starknet/src/messaging/mod.rs
```diff
@@ -0,0 +1,214 @@
...
+    /// * `from` - The block id from which (and including which) the messages are collected.
+    pub async fn collect_and_send_messages_to_l1(
+        &self,
+        from: u64,
+    ) -> DevnetResult<(Vec<MessageToL1>, u64)> {
+        let messaging = if let Some(m) = &self.messaging {
+            m
+        } else {
+            return Err(Error::MessagingError(MessagingError::NotConfigured));
+        };
```
FabijanC, line 134:
```suggestion
        let messaging = self.messaging_ref()?;
```

crates/starknet-server/tests/common/constants.rs
```diff
@@ -34,3 +34,6 @@ pub const CAIRO_1_PANICKING_CONTRACT_SIERRA_PATH: &str =
 // Contract which asserts the tx version
 pub const CAIRO_1_VERSION_ASSERTER_SIERRA_PATH: &str =
     "test_data/cairo1/version_asserter/version_asserter.sierra";
+
+// L1 L2 pre-registered addresses.
+pub const MESSAGING_L1_ALLOWED_CONTRACT: &str = "0x8359e4b0152ed5a731162d3c7b0d8d56edb165a0";
```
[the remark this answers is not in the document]
FabijanC: I remember seeing this was addressed. It would be good to have the commit link here, but I don't remember which one it was, so never mind, I'll just resolve this conversation.

contracts/generate_artifacts.sh
FabijanC: From my understanding, for general L1-L2 support, we only need the `MockStarknetMessaging` contract artifact. In devnet-py, those artifacts necessary for startup (i.e. necessary for users) were kept as a part of the git repo. However, those needed for testing (and potential example showcasing) were expected to be generated using a script.

I'd keep it the same way here. Currently [the documentation that a new user sees upon arriving to the repo](https://github.com/0xSpaceShard/starknet-devnet-rs#run) is wrong (incomplete): it is not sufficient to just come to the repo and do `cargo run`, you are also expected to run pre-build scripts.

contracts/README.md
FabijanC: The main `README.md` in the repo root should link to the one in `contracts/l1-l2-messaging`. E.g. by introducing a new `Development` subsection. Similar to e.g. `## Development - Testing`, there could be `## Development - L1-L2 (postman)`

contracts/l1-l2-messaging/README.md
```diff
@@ -0,0 +1,217 @@
...
+curl -H 'Content-Type: application/json' \
+    -d '{"paid_fee_on_l1": "0x123", "l2_contract_address": "0x03c80468c8fe2fd36fadf1b484136b4cd8a372f789e8aebcc6671e00101290a4", "l1_contract_address": "0xe7f1725E7734CE288F8367e1Bb143E90bb3F0512", "entry_point_selector": "0x00c73f681176fc7b3f9693986fd7b14581e8d540519e27400e88b8713932be01", "payload": ["0x1", "0x2"], "nonce": "0x1"}' \
+    http://127.0.0.1:5050/postman/send_message_to_l2
+```
+```json
+{
+    "transaction_hash": "0x7f5c523f47bc88fa21f86ec4aaac8bbad69dafb43ae7072319dcec4d5d40af9"
+}
+```
+The balance is now increased by 1, exactly has a message from l1 would have done.
```
FabijanC, line 204:
```suggestion
The balance is now increased by 1, exactly as a message from L1 would have done.
```

contracts/solidity/.env.anvil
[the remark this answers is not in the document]
FabijanC: The duplication regarding .env files has mostly been addressed, but why do we still have `l1-l2-messaging/cairo/.env` with a content that is a subset of `l1-l2-messaging/.env`? Consider this a question, not a change request.

contracts/cairo/README.md
```diff
@@ -0,0 +1,106 @@
...
+
+To ease the development, start the Devnet with a known seed (`--seed 42`) to use the pre-built account file.
+
+```bash
+# On a first terminal, run the Devnet with the option `--seed 42`.
+
+# Export variables to have starkli pre-configured.
+export STARKNET_RPC=http://127.0.0.1:5050
+export STARKNET_ACCOUNT=/path/account-0-42.json
+export STARKNET_PRIVATE_KEY=0xb137668388dbe9acdfa3bc734cc2c469
```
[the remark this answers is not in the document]
FabijanC: My comments regarding the instructions in README.md files weren't at all motivated by a lack of clarity. To the contrary, I found them clear to follow, and I had no problem having four terminals opened (ok, maybe someone else would mind this). But my biggest concern was having the same things (variable definitions and bash commands) defined in multiple places. If we ever change something, we will need to remember (or do clever enough ctrl+f searching around the repo) to address all of the occurrences.

crates/types/src/rpc/transactions/l1_handler_transaction.rs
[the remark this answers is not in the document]
glihm: I'm ok with that, I'll change into an `Err` and that's a good catch, it perhaps make more sense to check it only inside the `create_blockifier_transaction` because we know that the blockifier doesn't support other that version 0.

As all the fields are public, having a constructor would be too much arguments and members can easily changed. I'm more in favor to add the version check in `create_blockifier_transaction`.

Sounds good for you?

contracts/solidity/.env.anvil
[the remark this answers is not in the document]
glihm: addresses in the comment https://github.com/0xSpaceShard/starknet-devnet-rs/pull/234#discussion_r1406263072, you're totally right and I do think your take is better in this case.

crates/starknet-server/src/api/http/endpoints/postman.rs
glihm: Sorry didn't get you right.
No, it's not an error to flush if there is no messages. And now the behavior is modified to return empty arrays if the flush endpoint is queried and no message is available. :+1:

crates/starknet/src/starknet/add_l1_handler_transaction.rs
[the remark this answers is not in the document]
glihm: Newly fixed in 4d04967dbd996577b77c0a74c65a1c377184bb83, thank you again.

contracts/generate_artifacts.sh
glihm: That's a good point. The artifact is not re-introduced in 5208964828148554b1d5ef1102f60cc025605893.

crates/types/src/rpc/transaction_receipt.rs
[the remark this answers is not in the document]
FabijanC: Sorry, forgot to respond here. Really looks simplified now, but I'd like to see how you did it, i.e. a commit or two where I can see the exact changes.

contracts/README.md
ivpavici: fine by me! :)

crates/starknet/Cargo.toml
marioiordanov: its ok as it is

crates/starknet-server/tests/test_messaging.rs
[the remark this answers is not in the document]
glihm: implemented in d23c6405ce014dfe9705db0b2172f537bebd56d1.

contracts/README.md
glihm: Fixed in 483c52eefd1fb421a1f9396620c2d1a30fadd842 (and typo corrected in c00a45ddc5b47174afb65a2c9587b37d06294de7).

crates/starknet-server/src/api/http/models.rs
glihm: Fixed in 1341be6b69005e33eb62c9c534802b787b6802ed.

crates/types/src/rpc/transactions/l1_handler_transaction.rs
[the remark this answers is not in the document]
glihm: Fixed in bbedde4892b7ea48cccb8cd64c14dea47f906df3. At the end.. it's like other transaction types, the version is here, and it just need to be initialized correctly.
Perhaps the better way will be to add an enum if one day more version of L1HandlerTransaction are available.

crates/starknet/src/transactions.rs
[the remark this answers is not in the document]
glihm: Fixed in 01290817b75c047da870596d64b19a0aceb6deec.

FabijanC (approved):
I'm approving the changes after having modified (IMO improved) the tests a bit. I also added foundry installation as a CI step and mention it as a testing pre-requisite in the root `README.md`. All of the commits I pushed can be found at the bottom of the PR page. Let me know if you think I did something I shouldn't have.

glihm:
> I'm approving the changes after having modified (IMO improved) the tests a bit. I also added foundry installation as a CI step and mention it as a testing pre-requisite in the root `README.md`. All of the commits I pushed can be found at the bottom of the PR page. Let me know if you think I did something I shouldn't have.

You're all good @FabijanC very nice changes, thank you for that. The testing is more comprehensive.

rust-toolchain.toml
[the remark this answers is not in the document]
marioiordanov: Can't say anything about the error, probably an issue with the lock file. One of the reasons for having the cargo.lock is for not using any libs which bring breaking change to the project as it is the case with `clap`. It's not a big issue to upgrade to 1.70.0, although it's a kind of breaking change. I will resolve this conversation

mikiw (approved):

crates/starknet/src/messaging/ethereum.rs
FabijanC, line 1: There are still four TODOs in this file. Did you intend to resolve them before merging or did you leave them to be resolved in the future?

crates/starknet/src/messaging/ethereum.rs
```diff
@@ -371,11 +318,13 @@ impl EthereumMessaging {
                 e
             )))
         })?
-        .send()
-        .await
-        .map_err(|e| {
-            Error::MessagingError(MessagingError::EthersError(format!(
-                "Error deploying messaging contract: {}",
+            // Required by the new version of anvil, as default is no longer accepted.
+            .gas_price(1000000000)
```
FabijanC, line 322: How did you choose this value? How likely is it to fail?

glihm: I took the exact value that is used by anvil and hardhat as base fee.
I can put a bit more just to ensure that if this default value changes in the future it will not break.

crates/starknet/src/messaging/ethereum.rs
glihm: Let me review them to see which one are relevant or not. And changing them into a regular comment if there is something important to keep in mind.

crates/starknet/src/messaging/ethereum.rs
glihm: Comment added in 9cc2be46c51128155506d036c1d53621f281e79d and I multiplied by 20 to have a safe value as the default value is not expected to change everyday by anvil/hardhat releases.

PR #235, opened 2023-10-31
Fix block timestamp

marioiordanov:
## Usage related changes

Before, latest block timestamp was very close in value to the last block generated, or 0 in the case of being the first block.
This was happening, because when creating a new block it was generated from the pending block, which timestamp property was generated after creating the latest block.

This PR resolves that and removes the relation between pending block timestamp and new block timestamp, by assigning the block timestamp, when the block is created. Not when the last block is generated

- Remove code that sets value regarding timestamp in`update_block_context` method.
- Updated `generate_new_block` to set the timestamp of the block with UNIX time.

## Checklist:

- [x] Documented the changes
- [x] Updated the tests

### DIFF
--- a/crates/starknet/src/starknet/mod.rs
+++ b/crates/starknet/src/starknet/mod.rs
@@ -1207,8 +1207,8 @@ mod tests {
         let now = Starknet::get_unix_timestamp_as_seconds();
 
         let block_timestamp = starknet.get_latest_block().unwrap().header.timestamp;
-        // check if the pending_block_timestamp is less than the block_timestamp, by number of sleep seconds
-        // because the timeline of events is this:
+        // check if the pending_block_timestamp is less than the block_timestamp, by number of sleep
+        // seconds because the timeline of events is this:
         // ----(pending block timestamp)----(sleep)----(new block timestamp)
         assert!(pending_block_timestamp.0 + sleep_duration_secs <= block_timestamp.0);
         // check if now is close to the block_timestamp
--- a/crates/starknet/src/starknet/mod.rs
+++ b/crates/starknet/src/starknet/mod.rs
@@ -1211,7 +1211,5 @@ mod tests {
         // by number of sleep seconds because the timeline of events is this:
         // ----(pending block timestamp)----(sleep)----(new block timestamp)
         assert!(pending_block_timestamp.0 + sleep_duration_secs <= block_timestamp.0);
-        // check if now is close to the block_timestamp
-        assert!(now - block_timestamp.0 <= 1);
     }
 }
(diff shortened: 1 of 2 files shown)


crates/starknet/src/starknet/mod.rs
```diff
@@ -1184,4 +1187,29 @@ mod tests {
...
     }
+    #[test]
+    fn check_timestamp_of_newly_generated_block() {
+        let config = starknet_config_for_test();
+        let mut starknet = Starknet::new(&config).unwrap();
+
+        Starknet::update_block_context(&mut starknet.block_context);
+        starknet.generate_pending_block().unwrap();
+        starknet.blocks.pending_block.set_timestamp(BlockTimestamp(
+            std::time::SystemTime::now().duration_since(std::time::UNIX_EPOCH).unwrap().as_secs(),
```
FabijanC, line 1198: Since the same code for getting current time is used in multiple places, why not extract it? Maybe also use the same function above in L182?

crates/starknet/src/starknet/mod.rs
```diff
@@ -1184,4 +1187,29 @@ mod tests {
...
+        let sleep_duration_secs = 5;
+        thread::sleep(Duration::from_secs(sleep_duration_secs));
+        starknet.generate_new_block(StateDiff::default()).unwrap();
+        let now =
+            std::time::SystemTime::now().duration_since(std::time::UNIX_EPOCH).unwrap().as_secs();
+
+        let block_timestamp = starknet.get_latest_block().unwrap().header.timestamp;
+        // check if the pending_block_timestamp is less than the block_timestamp, by at least the
+        // sleep duration - 1
+        assert!(pending_block_timestamp.0 + sleep_duration_secs - 1 < block_timestamp.0);
```
FabijanC, line 1211: Why not use `<=` in both cases? Also, I think the comments should explain where the `- 1` comes from in these assertions. In fact, I think the clearest way would be to introduce a function like this:
```rust
    /// Uses EPSILON=1
    fn assert_timestamps_almost_equal(t1: u64, t2: u64) {
        assert!(t1.abs_diff(t2) <= 1);
    }
```
and use for both assertions, in the first case it being:
```rust
assert_timestamps_almost_equal(
  pending_block_timestamp.0 + sleep_duration_secs,
  block_timestamp.0,
)
```

FabijanC (changes requested):

crates/starknet/src/starknet/mod.rs
marioiordanov: extracted in [9478662](https://github.com/0xSpaceShard/starknet-devnet-rs/pull/235/commits/947866291971f3ce316e3cf2b03cfc068bb9f032)

crates/starknet/src/starknet/mod.rs
marioiordanov: removed the check with -1 and updated the comment. Didn't introduce the suggested function, because it was used only for 1 check

FabijanC (approved):

Issue #237, opened 2023-11-01
Improve the dump and load feature


mikiw:
The dump and load feature needs to be extended to handle HTTP endpoint calls like create_block, flush, send_message_to_l2, consume_message_from_l2

mikiw:
Since the restart feature was added consider adding test: dump, restart, load.

PR #238, opened 2023-11-02
Add restarting functionality

### LINKED ISSUE #232: No restarting functionality
This is one of the features present in devnet-py: https://0xspaceshard.github.io/starknet-devnet/docs/guide/restart

FabijanC:
## Usage related changes

- Closes #232 

- Refactor BackgroundDevnet CLI parsing to allow custom arguments in array/slice form, but also to use default values for those needed and not provided
- Remove TODOs from old tests
  - By changing how max_fee is specified in one test, its tx hash was changed, so I took the chance to remove some of the unnecessary constants that are used in just one place.
- Introduce `devnet.get_balance(address)` and use it in old and new tests
- Move `send_ctrl_c_signal` to utils and make it more general (accept any process, not necessarily BackgroundDevnet)

## Checklist:

- [x] Documented the changes
- [x] Updated the tests

### DIFF
--- a/crates/starknet-server/tests/common/devnet.rs
+++ b/crates/starknet-server/tests/common/devnet.rs
@@ -85,12 +85,13 @@ impl BackgroundDevnet {
     /// Takes user specified args and adds default values for args that are missing
     fn add_default_args<'a>(user_args: &[&'a str]) -> Vec<&'a str> {
         let mut substituted_args: Vec<&'a str> = vec![];
-        for (arg, default_value) in CLI_MAP.iter() {
-            substituted_args.push(arg);
+        for (arg_name, default_value) in CLI_MAP.iter() {
+            substituted_args.push(arg_name);
 
             let value = user_args
                 .iter()
-                .position(|arg_candidate| arg_candidate == arg)
+                .position(|arg_candidate| arg_candidate == arg_name)
+                // arg value comes after name
                 .map(|pos| user_args[pos + 1])
                 .unwrap_or(default_value);
             substituted_args.push(value);
--- a/crates/starknet-server/tests/common/devnet.rs
+++ b/crates/starknet-server/tests/common/devnet.rs
@@ -66,7 +66,7 @@ fn get_free_port() -> Result<u16, TestError> {
 }
 
 lazy_static! {
-    static ref CLI_MAP: HashMap<&'static str, String> = HashMap::from([
+    static ref DEFAULT_CLI_MAP: HashMap<&'static str, String> = HashMap::from([
         ("--seed", SEED.to_string()),
         ("--accounts", ACCOUNTS.to_string()),
         ("--initial-balance", PREDEPLOYED_ACCOUNT_INITIAL_BALANCE.to_string()),
@@ -82,22 +82,29 @@ impl BackgroundDevnet {
         BackgroundDevnet::spawn_with_additional_args(&[]).await
     }
 
-    /// Takes user specified args and adds default values for args that are missing
-    fn add_default_args<'a>(user_args: &[&'a str]) -> Vec<&'a str> {
-        let mut substituted_args: Vec<&'a str> = vec![];
-        for (arg_name, default_value) in CLI_MAP.iter() {
-            substituted_args.push(arg_name);
-
-            let value = user_args
-                .iter()
-                .position(|arg_candidate| arg_candidate == arg_name)
-                // arg value comes after name
-                .map(|pos| user_args[pos + 1])
-                .unwrap_or(default_value);
-            substituted_args.push(value);
+    /// Takes specified args and adds default values for args that are missing
+    fn add_default_args<'a>(specified_args: &[&'a str]) -> Vec<&'a str> {
+        let mut specified_args_vec: Vec<&str> = specified_args.to_vec();
+        let mut final_args: Vec<&str> = vec![];
+        // iterate through default args, and remove from specified when found
+        // that way in the end we can just extend the final vec with the remaining specified args
+        for (arg_name, default_value) in DEFAULT_CLI_MAP.iter() {
+            let value =
+                match specified_args_vec.iter().position(|arg_candidate| arg_candidate == arg_name)
+                {
+                    Some(pos) => {
+                        // arg value comes after name
+                        specified_args_vec.remove(pos);
+                        specified_args_vec.remove(pos)
+                    }
+                    None => default_value,
+                };
+            final_args.push(arg_name);
+            final_args.push(value);
         }
 
-        substituted_args
+        final_args.extend(specified_args_vec);
+        final_args
     }
 
     pub(crate) async fn spawn_with_additional_args(args: &[&str]) -> Result<Self, TestError> {
--- a/crates/starknet-server/tests/common/devnet.rs
+++ b/crates/starknet-server/tests/common/devnet.rs
@@ -9,10 +9,14 @@ use hyper::http::request;
 use hyper::{Body, Client, Response, StatusCode, Uri};
 use lazy_static::lazy_static;
 use serde_json::json;
-use starknet_rs_core::types::FieldElement;
+use starknet_core::constants::ERC20_CONTRACT_ADDRESS;
+use starknet_rs_core::types::{BlockId, BlockTag, FieldElement, FunctionCall};
+use starknet_rs_core::utils::get_selector_from_name;
 use starknet_rs_providers::jsonrpc::HttpTransport;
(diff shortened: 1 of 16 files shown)


crates/starknet-server/tests/get_transaction_by_hash.rs
```diff
@@ -133,10 +126,8 @@ mod get_transaction_by_hash_integration_tests {
             starknet_rs_core::types::DeclareTransaction::V2(declare_v2),
         ) = result
         {
-            assert_eq!(
-                declare_v2.transaction_hash,
-                FieldElement::from_hex_be(DECLARE_V2_TRANSACTION_HASH).unwrap()
-            );
+            let expected = "0x013070e9ff8b554254fb2d577ea75f0ed925947902893a23b8c738f8fbe5e122";
```
mikiw, line 129: Can we use const at the beginning of tests like it was before?

crates/starknet-server/tests/get_transaction_by_hash.rs
```diff
@@ -176,10 +167,8 @@ mod get_transaction_by_hash_integration_tests {
             .unwrap();
 
         if let starknet_rs_core::types::Transaction::DeployAccount(deploy) = result {
-            assert_eq!(
-                deploy.transaction_hash,
-                FieldElement::from_hex_be(DEPLOY_ACCOUNT_TRANSACTION_HASH).unwrap()
-            );
+            let expected = "0x02b4b37075273f836aa055c7a53e4e2635abcbd776ebef2ab1b74abd7235ac06";
```
mikiw, line 170: Can we use const at the beginning of tests like it was before?

crates/starknet-server/tests/test_call.rs
```diff
@@ -13,6 +13,7 @@ mod call {
     use crate::common::devnet::BackgroundDevnet;
 
     #[tokio::test]
+    /// This test doesn't rely on devnet.get_balance because it's not supposed to call ERC20
```
mikiw, line 16: I don't get this comment. Shouldn't this be "This test doesn't rely on balanceOf call"?

crates/starknet-server/tests/test_restart.rs
```diff
@@ -0,0 +1,255 @@
...
+        // assert killing the process can still dump devnet
+        send_ctrl_c_signal(&devnet.process).await;
+        std::thread::sleep(std::time::Duration::from_secs(1)); // sleep to allow dump
+        assert!(Path::new(dump_file_name).exists());
+
+        remove_file(dump_file_name);
+    }
+
+    #[tokio::test]
+    async fn assert_load_not_affecting_restart() {
```
mikiw, line 219: We could also add test with dump, restart, load, check if tx is there but this seems to be more dump and load test than restart test. I can also add it by my own during https://github.com/0xSpaceShard/starknet-devnet-rs/issues/237

crates/starknet-server/tests/test_restart.rs
```diff
@@ -0,0 +1,255 @@
...
+    async fn assert_tx_not_present_after_restart() {
+        let devnet = BackgroundDevnet::spawn().await.unwrap();
+
+        // generate dummy tx
+        let mint_hash = devnet.mint(FieldElement::ONE, 100).await;
+        assert!(devnet.json_rpc_client.get_transaction_by_hash(mint_hash).await.is_ok());
+
+        let restart_resp = devnet.restart().await.unwrap();
+        assert_eq!(restart_resp.status(), StatusCode::OK);
+
```
mikiw, line 43: We could also check if there is no block after restart.

mikiw (changes requested):

crates/starknet-server/tests/get_transaction_by_hash.rs
FabijanC: These are very specific constants (like, the hash of a transaction), and highly unlikely that they will be used in more than one place. If in the future we really have the constant used in two places, I'm ok with extracting it. Also, it uses less lines like this and I think the clarity is the same.

crates/starknet-server/tests/get_transaction_by_hash.rs
FabijanC: See: https://github.com/0xSpaceShard/starknet-devnet-rs/pull/238#discussion_r1389116033

crates/starknet-server/tests/test_call.rs
FabijanC: I only added this comment to explain why I'm not using the `get_balance` util method, because someone might think it could be used here.

Let's look at what this test case really does. As the name says, it's `calling_method_of_undeployed_contract`. Whoever wrote this test case (looks like it was me), decided to test how devnet behaves on calling an undeployed contract by trying to simulate the calling of ERC20, but with an invalid address (i.e. an undeployed address, i.e. an address where nothing has been deployed).

If you look at the tests in this file (they are all testing `starknet_call`):
- the first one is calling, but with a wrong address
- the second one is calling, but with a wrong method (non-existent method selector)
- the third one is calling the correct address with a correct method (happy path)

crates/starknet-server/tests/test_restart.rs
FabijanC: Yes, looks more like a dump/load test, I don't know if we need it.

crates/starknet-server/tests/common/devnet.rs
```diff
@@ -64,6 +69,15 @@ fn get_free_port() -> Result<u16, TestError> {
     Err(TestError::NoFreePorts)
 }
 
+lazy_static! {
```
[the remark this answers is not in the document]
FabijanC: Without lazy_static, I get:
```
cannot call non-const fn `<HashMap<&str, std::string::String> as From<[(&str, std::string::String); 4]>>::from` in statics
calls in statics are limited to constant functions, tuple structs and tuple variants
consider wrapping this expression in `Lazy::new(|| ...)` from the `once_cell` crate
```

The alternative would be to define this map in the function where it is used, but I think this is clearer.

crates/starknet-server/tests/common/devnet.rs
```diff
@@ -72,6 +86,33 @@ impl BackgroundDevnet {
         BackgroundDevnet::spawn_with_additional_args(&[]).await
     }
 
+    /// Takes specified args and adds default values for args that are missing
```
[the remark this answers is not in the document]
FabijanC: Otherwise I cannot use BackgroundDevnet for testing a provided `--initial-balance`. How else would I test a BackgroundDevnet with a custom port, or chain-id, or accounts number? Without this intervention, using
```rust
BackgroundDevnet::spawn_with_additional_args(&["--initial-balance", "100"])
```
would result in a command that internally uses
```
cargo run --release -- --initial-balance DEFAULT_VALUE --initial-balance 100 ...
```

In devnet-py, arg parser allowed this duplicate argument specification (it would simply use the last instance of the specifications), but here with Clap it's not allowed (it errors), so we have to make sure each CLI arg is supplied at most once.

crates/starknet-server/tests/test_restart.rs
FabijanC: Added in https://github.com/0xSpaceShard/starknet-devnet-rs/pull/238/commits/8c566b5e6b2ac654f66e90baedb4096c1383969d

crates/starknet-server/tests/common/devnet.rs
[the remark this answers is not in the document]
mikiw: ok got it

crates/starknet-server/tests/test_call.rs
mikiw: ok got it now

crates/starknet-server/tests/get_transaction_by_hash.rs
mikiw: I'm aware that it will be not used elsewhere and it's not the point. Maybe I'll rephrase it. Currently, we have 2 syntaxes in tests one is the use of pub const that has names like `DECLARE_V1_TRANSACTION_HASH` defined at the top and you introduced new syntax like expected ="0x1" inside test.

Let's use one of it, `pub const` on the top or `expected = "0x1"`. I really don't care which one I just care about cohesion.

FabijanC: Addressed in https://github.com/0xSpaceShard/starknet-devnet-rs/pull/238/commits/f3aa8c1b805d592115c1a0f1fd8c92567cae80c8

mikiw (approved):

PR #253, opened 2023-11-17
Testing improvements

### LINKED ISSUE #205: Improve dump and load testing
These tests usually dump devnet into a file. We need to come up with new file names to ensure each test has a unique name and doesn't affect other tests which may be run in parallel. For this purpose we could use [NamedTempFile](https://docs.rs/tempfile/latest/tempfile/struct.NamedTempFile.html) or something similar.

This would also remove the need to remember to delete the file in the end of each test, also leading to less code.

cc: @mikiw

### LINKED ISSUE #243: starknet_config_for_test and Default for StarknetConfig can be unified
First idea: use values from `starknet_config_for_test`, but keep `impl Default for StarknetConfig`

The used seeds are not the same so some expected hash and address values might change, but this can be substituted with the new expected values.

FabijanC:
## Usage related changes

None

Introduces some refactoring to improve the overall testing experience

- Closes #243
- Renames `devnet.rs` to `background_devnet.rs`
  - This has been bugging me for some time - having a file called devnet.rs inside the devnet repository is just wrong.
- Introduces `UniqueAutoDeletableFile`
  - Closes #205
  - Unlike [NamedTempFile](https://docs.rs/tempfile/latest/tempfile/struct.NamedTempFile.html), it doesn't create the file on struct construction, so it is more suitable for our use case.
  - Useful in dumping/loading tests.
  - Prevents name collisions - we no longer have to come up with unique dump artifact names.
  - Automatically deletes the underlying file when the variable is dropped - no need to remember to delete.

## Checklist:

- [x] Documented the changes
- [x] Updated the tests


crates/starknet-server/tests/common/utils.rs
```diff
@@ -139,3 +140,47 @@ pub async fn send_ctrl_c_signal(process: &Child) {
...
+
+#[cfg(test)]
+mod test_unique_auto_deletable_file {
+    use std::path::Path;
+
+    use super::UniqueAutoDeletableFile;
+
+    #[test]
+    fn foo() {
+        let file = UniqueAutoDeletableFile::new("foo");
```
mikiw, line 176: I think that one test should be added with 2 instances of `UniqueAutoDeletableFile` like:
```
let first_foo = UniqueAutoDeletableFile::new("foo");
let second_foo = UniqueAutoDeletableFile::new("foo");
```

to prove that collisions will be prevented

mikiw (changes requested):

crates/starknet-server/tests/common/utils.rs
FabijanC: Thanks for the suggestion. I added new tests in https://github.com/0xSpaceShard/starknet-devnet-rs/pull/253/commits/ba8de5bcad4f1fd4906e800203cce401b40b774e

mikiw: I would remove that loop, it's basically a test of `rand::thread_rng` it doesn't make sense in my opinion to test that in our code base

mikiw (approved):

PR #264, opened 2023-11-27
Add improved timestamp contract

mikiw:

- improved timestamp contract and tests

## Checklist:

- [x] Documented the changes
- [x] Updated the tests


crates/starknet-server/tests/test_advancing_time.rs
```diff
@@ -72,22 +67,177 @@ mod advancing_time_tests {
...
+        contract_address: FieldElement,
+    ) -> FieldElement {
+        let call_current_timestamp = FunctionCall {
+            contract_address,
+            entry_point_selector: get_selector_from_name("get_timestamp").unwrap(),
+            calldata: vec![],
+        };
+        devnet
+            .json_rpc_client
+            .call(call_current_timestamp.clone(), BlockId::Tag(BlockTag::Latest))
```
FabijanC, line 89: unnecessary clone

crates/starknet-server/tests/test_advancing_time.rs
```diff
@@ -72,22 +67,177 @@ mod advancing_time_tests {
...
+            .await
+            .unwrap()[0]
```
FabijanC, line 91: Before we take the presumably only element with `[0]`, IMO it would be good to assert that the received iterable really has a length of one.

crates/starknet-server/tests/test_advancing_time.rs
```diff
@@ -72,22 +67,177 @@ mod advancing_time_tests {
...
+
+        // set time in past
+        let past_time = 1;
+        let set_time_body = Body::from(json!({ "time": past_time }).to_string());
+        let resp_set_time = devnet.post_json("/set_time".into(), set_time_body).await.unwrap();
+        let resp_body_set_time = get_json_body(resp_set_time).await;
+        assert_eq!(resp_body_set_time["block_timestamp"], past_time);
+
+        // mine block
+        devnet.mint(DUMMY_ADDRESS, DUMMY_AMOUNT).await;
```
FabijanC, line 108: We're really relying a lot on this minting. I suggest that for this test file we create a function called `mine_new_block`, which will still mint, but at least we will have it abstracted, and also there will be no need to write these `// mine block` comments as it will be self-explanatory.

crates/starknet-server/tests/test_advancing_time.rs
```diff
@@ -72,22 +67,177 @@ mod advancing_time_tests {
...
+
+        // check if timestamp is greater/equal
+        let current_timestamp = get_current_timestamp(&devnet, new_contract_address).await;
+        assert_ge_with_buffer(current_timestamp.to_string().parse::<u64>().ok(), Some(past_time));
```
FabijanC, line 112: Here and in other places, wouldn't it be better to let `get_current_timestamp` do the conversion from `FieldElement` to `Option<u64>`, rather than doing it in multiple places.

crates/starknet-server/tests/test_advancing_time.rs
```diff
@@ -72,22 +67,177 @@ mod advancing_time_tests {
...
             FieldElement::ZERO,
             declaration_result.class_hash,
             &starknet_rs_core::utils::UdcUniqueness::NotUnique,
             &[],
+        )
+    }
+
+    pub async fn get_current_timestamp(
+        devnet: &BackgroundDevnet,
+        contract_address: FieldElement,
```
FabijanC, line 80: It would be clearer to call this `timestamp_contract_address`.

crates/starknet-server/tests/test_advancing_time.rs
```diff
@@ -72,22 +67,177 @@ mod advancing_time_tests {
...
+            .call(call_current_timestamp.clone(), BlockId::Tag(BlockTag::Latest))
+            .await
+            .unwrap()[0]
+    }
+
+    #[tokio::test]
+    async fn timestamp_syscall_set_in_past() {
+        let devnet: BackgroundDevnet =
+            BackgroundDevnet::spawn().await.expect("Could not start Devnet");
+        let new_contract_address = setup_timestamp_contract(&devnet).await;
```
FabijanC, line 98: Here and in other places, it would really be more indicative if the variable was not called `new_contract_address` but `timestamp_contract_address`.

crates/starknet-server/tests/test_advancing_time.rs
```diff
@@ -133,7 +283,7 @@ mod advancing_time_tests {
         // set time and assert if it's greater/equal than future_time, check if it's inside buffer
         // limit
         let now = get_unix_timestamp_as_seconds();
-        let future_time = now + 100;
+        let future_time = now + 1000;
```
FabijanC, line 286: Why the increase?

crates/starknet-server/tests/test_advancing_time.rs
```diff
@@ -72,22 +67,177 @@ mod advancing_time_tests {
...
+            BackgroundDevnet::spawn().await.expect("Could not start Devnet");
+        let new_contract_address = setup_timestamp_contract(&devnet).await;
+
+        // increase time
+        let increase_time: u64 = 1000;
+        let increase_time_body = Body::from(json!({ "time": increase_time }).to_string());
+        devnet.post_json("/increase_time".into(), increase_time_body).await.unwrap();
+
+        // check if timestamp is greater/equal
+        let mut current_timestamp = get_current_timestamp(&devnet, new_contract_address).await;
```
FabijanC, line 150: Instead of having a mutable `current_timestamp`, why not have to different variables, which can then also be asserted as in the next test (`timestamp_syscall_contract_constructor`) - that the second one is greater.

FabijanC (changes requested):

crates/starknet-server/tests/test_advancing_time.rs
mikiw: Explained on the call

mikiw: comment added in 5b8dba1ad980cd3310fd4a14ed78eb636e12b8ea

crates/starknet-server/tests/test_advancing_time.rs
mikiw: fixed in 5b8dba1ad980cd3310fd4a14ed78eb636e12b8ea

crates/starknet-server/tests/test_advancing_time.rs
mikiw: true, renamed in 5b8dba1ad980cd3310fd4a14ed78eb636e12b8ea

crates/starknet-server/tests/test_advancing_time.rs
mikiw: renamed in 5b8dba1ad980cd3310fd4a14ed78eb636e12b8ea

crates/starknet-server/tests/test_advancing_time.rs
mikiw: fixed in 5b8dba1ad980cd3310fd4a14ed78eb636e12b8ea

crates/starknet-server/tests/test_advancing_time.rs
mikiw: right, it was needed before but later I moved it to a new function

fixed in 5b8dba1ad980cd3310fd4a14ed78eb636e12b8ea

crates/starknet-server/tests/test_advancing_time.rs
mikiw: fixed in 98d123dbee60863f2483f5fa658d9bc570c5e404

crates/starknet-server/tests/test_advancing_time.rs
mikiw: fixed in f7d38a3368700e4f76b2df88193217df15e45cb9

crates/starknet-server/tests/test_advancing_time.rs
FabijanC: From what I see, the `.clone()` is still here.

mikiw: sorry for that, probably I didn't save file, fixed in d6bd56b39308863cd77bfbe777a394cc40ab1e57

FabijanC (approved):

PR #257, opened 2023-11-22
0.5.1 spec

marioiordanov:
## Usage related changes

0.5.1 spec

## Development related changes

- Added `.vscode/` folder to .`gitignore`
- Added 0.5.1 files
- Changed `state_diff` in `TransactionTrace` to be optional
- Added `CallType::Delegate`

## Checklist:

- [ ] Applied formatting - `./scripts/format.sh`
- [ ] No linter errors - `./scripts/clippy_check.sh`
- [ ] Performed code self-review
- [ ] Rebased to the last commit of the target branch (or merged it into my branch)
- [ ] Documented the changes
- [ ] Linked the issues which this PR resolves
- [ ] Checked the TODO section in README.md if this PR resolves it
- [ ] Updated the tests
- [ ] All tests are passing - `cargo test`

PR #266, opened 2023-11-29
Add json rpc starknet_traceTransaction

mikiw:
## Usage related changes

- Add of starknet_traceTransaction json rpc endpoint

- refactor of `try_from_call_info()`

## Checklist:

- [x] Documented the changes
- [x] Updated the tests


mikiw:
TODO on Monday: Add declare, deploy + tests. Refactor common code that it's also in simulate function.

crates/starknet-server/tests/test_dump_and_load.rs
```diff
@@ -1,6 +1,5 @@
 pub mod common;
 
-// Important! Use unique file names for dump files, tests can be run in parallel.
```
mikiw, line 3: this comment is not needed anymore

crates/types/src/rpc/transactions.rs
```diff
@@ -435,11 +435,11 @@ pub struct SimulatedTransaction {
 
 impl FunctionInvocation {
     pub fn try_from_call_info(
-        mut call_info: blockifier::execution::call_info::CallInfo,
+        call_info: &blockifier::execution::call_info::CallInfo,
```
FabijanC, line 438: Why was this needed?

crates/starknet/src/starknet/mod.rs
```diff
@@ -705,6 +706,107 @@ impl Starknet {
...
+                    }
+                }
+            }
+        })
+    }
+
+    fn get_call_info_invocation(
+        call_info_invocation: &Option<CallInfo>,
+        address_to_class_hash_map: &HashMap<ContractAddress, ClassHash>,
+    ) -> DevnetResult<Option<FunctionInvocation>> {
```
FabijanC, line 746: Wrapping an `Option` in a `(Devnet)Result` looks wrong. Also, this implementation looks like it could benefit from methods of `Option`.

crates/starknet/src/starknet/mod.rs
```diff
@@ -705,6 +706,107 @@ impl Starknet {
...
+    ) -> DevnetResult<TransactionTrace> {
+        let transaction_to_map =
+            self.transactions.get(&transaction_hash).ok_or(Error::NoTransaction)?;
+
+        let mut state =
+            self.get_state_at(&BlockId::Tag(starknet_rs_core::types::BlockTag::Latest))?.clone();
+
+        let state_diff: ThinStateDiff = state.extract_state_diff_from_pending_state()?.into();
+        let state_diff =
+            if state_diff == ThinStateDiff::default() { None } else { Some(state_diff) };
```
FabijanC, line 766: This is already the second time that we are introducing this hacky check.
It's been discussed in [this conversation](https://github.com/0xSpaceShard/starknet-devnet-rs/pull/257/files#r1401714563) with @marioiordanov.

Maybe more logic could be extracted to a common function used by `get_transaction_trace` and `simulate_transactions`?

FabijanC (changes requested):

marioiordanov (changes requested):
In starknet-server/test_data/spec/0.5.1/starknet_trace_api_openrpc.json the method starknet_traceTransaction is missing, because until this PR there was no need for that. Please add it and check if serialization/deserialization works for it, by running `spec_reader/mod.rs` tests

crates/types/src/rpc/transactions.rs
marioiordanov: `starknet_traceTransaction` is invoked for already executed transactions. They are stored in collection and getting them from it returns a reference. The reason for changing the signature of the method is that `blockifier::execution::call_info::CallInfo` doesnt implement Copy/Clone and using it with an object that is a reference struct results in `moving out of shared reference` error.

crates/types/src/rpc/transactions.rs
```diff
@@ -450,23 +450,28 @@ impl FunctionInvocation {
...
-
-        let messages: Vec<OrderedMessageToL1> = call_info
+        let mut messages: Vec<OrderedMessageToL1> = call_info
             .execution
             .l2_to_l1_messages
-            .into_iter()
-            .map(|msg| OrderedMessageToL1::new(msg, call_info.call.caller_address.into()))
+            .iter()
+            .map(|msg| {
+                OrderedMessageToL1::new(
```
marioiordanov, line 458: `OrderedMessageToL1::new` method is used only in this place. Why not refactoring it to accept `&OrderedL2ToL1Message` instead of `OrderedL2ToL1Message`.

crates/starknet/src/starknet/mod.rs
mikiw: @marioiordanov? I was just refactoring your code :)

FabijanC: It might just be my Rust inexperience and this pattern is considered normal, but it does look suspicious.

crates/starknet/src/starknet/mod.rs
mikiw: fixed in 2fac26806808f07a966160c7249bb15458169ce3

mikiw: Ok not fixed and will not be fixed I guess...

I needed to revert that in f888865b0350b62903ce7d246c2437c806b39cee. The problem is that this refactor will cause extra clones of that state :(

Additionally, it gave me a headache (×_×;）

mikiw:
> In starknet-server/test_data/spec/0.5.1/starknet_trace_api_openrpc.json the method starknet_traceTransaction is missing, because until this PR there was no need for that. Please add it and check if serialization/deserialization works for it, by running `spec_reader/mod.rs` tests

Added in b30756459ad5516382be568bbf2e866c2f13e5dd

crates/starknet-server/src/api/json_rpc/endpoints.rs
```diff
@@ -374,4 +374,17 @@ impl JsonRpcHandler {
...
+    /// starknet_traceTransaction
+    pub(crate) async fn get_trace_transaction(
+        &self,
+        transaction_hash: TransactionHash,
+    ) -> StrictRpcResult {
+        let starknet = self.api.starknet.read().await;
+        match starknet.get_transaction_trace_by_hash(transaction_hash) {
+            Ok(result) => Ok(StarknetResponse::TraceTransaction(result)),
+            Err(Error::NoTransaction) => Err(ApiError::TransactionNotFound),
+            Err(err) => Err(err.into()),
```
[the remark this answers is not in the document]
mikiw: So I added `NO_TRACE_AVAILABLE` but I'm not sure why are still not handling `INVALID_TXN_HASH`?

crates/starknet/src/starknet/mod.rs
marioiordanov: > @marioiordanov? I was just refactoring your code :)

What should be my expected reply

FabijanC: > What should be my expected reply

One of these:
- telling us of a magical way to refactor this not to use Results wrapped in Options
- telling us that this is a common pattern and we lack experience

marioiordanov: Ahh, okay. ```rust
get_call_info_invocation(call_info_invocation: &CallInfo,
        address_to_class_hash_map: &HashMap<ContractAddress, ClassHash>)-> DevnetResult<FunctionInvocation>``` . By removing the option from call_info_invocation it is removed from the result.
        
 Before executing get_call_info_invocation, the match statement will be in the form:
 ```rust
if let Some(ref x) = transaction_to_map.execution_info.validate_call_info {
            Some(FunctionInvocation::try_from_call_info(x, address_to_class_hash_map)?)
        } else {
            None
        };
 ```

crates/starknet-server/src/api/json_rpc/endpoints.rs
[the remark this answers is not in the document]
mikiw: ok for now I added something like this 3b2692f0ebf0dde4c7b852f056c372ef50e408ad is that ok?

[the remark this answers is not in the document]
mikiw: I would leave it as it is for now, once there will be reject support or better specification we can change it again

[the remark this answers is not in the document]
FabijanC: Ok, since we are never rejecting, and our txs are never RECEIVED, we can resolve this.

FabijanC (approved):
Giving my approval, but let's resolve the remaining refactoring request

mikiw:
L1hanlder transactions support will be added in https://github.com/0xSpaceShard/starknet-devnet-rs/issues/280

crates/starknet/src/starknet/mod.rs
marioiordanov: For me it doesn't look good to be Result<Option<T>> but If there is no other way I will resolve this conversation. Also creating an issue for this refactoring doesn't make sense, because this part was refactored already in this PR.

marioiordanov (approved):
Approving, but lets refer to this comment https://github.com/0xSpaceShard/starknet-devnet-rs/pull/266/files#r1420574368

PR #284, opened 2023-12-12
Add starknet_traceBlockTransactions 

mikiw:
## Checklist:

- [x] Documented the changes
- [x] Updated the tests


crates/starknet-server/tests/test_trace.rs
```diff
@@ -189,4 +188,54 @@ mod trace_tests {
...
+        // this test also needs to be updated
+        let block_traces = &devnet
+            .send_custom_rpc("starknet_traceBlockTransactions", json!({ "block_id": "latest" }))
+            .await["result"];
+
+        // assert transaction hash
+        assert_eq!(
+            mint_tx_hash,
+            FieldElement::from_hex_be(
+                block_traces["traces"][0]["transaction_hash"].as_str().unwrap()
```
FabijanC, line 208: Is len of `block_traces["traces"]` supposed to be one? If so, I would add a length assertion before all the assertions, and extract `block_traces["traces"][0]` into a `trace` variable to reduce duplication.

FabijanC (commented):
Looks good, just one question (potential change request) in the tests.

crates/starknet-server/tests/test_trace.rs
mikiw: addressed in cdfa97f922e9acc47385258a71ad42a3195bfae9

FabijanC: If you already deserialize it with
```
serde_json::from_value::<Vec<BlockTransactionTrace>>(block_traces.clone(),)
```
why not later use `validate_invocation`, `execute_invocation` and `fee_transfer_invocation` of that deserialized object that is only used for length assertion?

Also, no need to call the variable `traces` if it's just one `trace` (the one at index 0)

mikiw: that would require if let type match which would increase code size

FabijanC: I see. but why can't the length be asserted on the object whose property [0] you are taking? I.e. without deserialization?

mikiw: because it's `&Value` type

FabijanC: ok then

FabijanC (approved):

PR #290, opened 2023-12-13
Add state-archive flag

mikiw:
[skip ci]

## Checklist:

- [x] Documented the changes
- [x] Updated the tests


crates/starknet/src/starknet/starknet_config.rs
```diff
@@ -57,6 +58,7 @@ impl Default for StarknetConfig {
             dump_on: None,
             dump_path: None,
             re_execute_on_init: true,
+            state_archive: false,
```
ivpavici, line 61: discussing with some projects that use devnet-rs, for now it seems the default can be `false`

crates/starknet/src/starknet/get_class_impls.rs
```diff
@@ -129,9 +135,23 @@ mod tests {
...
+        let (mut starknet, account) = setup(Some(100000000), StateArchiveCapacity::None);
+
+        starknet.generate_new_block(StateDiff::default(), None).unwrap();
+        starknet.generate_pending_block().unwrap();
+
+        let block_number = starknet.get_latest_block().unwrap().block_number();
+        let block_id = BlockId::Number(block_number.0);
+
+        let class_hash = starknet.get_class_hash_at(block_id, account.account_address);
+        assert!(class_hash.is_err());
```
FabijanC, line 149: Can you assert the error type? So that we're sure it's not failing with something unforeseen

crates/starknet/src/starknet/starknet_config.rs
```diff
@@ -57,6 +66,7 @@ impl Default for StarknetConfig {
             dump_on: None,
             dump_path: None,
             re_execute_on_init: true,
+            state_archive: StateArchiveCapacity::None,
```
FabijanC, line 69: Can you instead do `StateArchiveCapacity::default()`? If we ever change it, it would be good that it's changed in just one place. I think if you derive Default, it will use the first enum option as the default (unless specified otherwise) - which is what we need (`none`).

README.md
```diff
@@ -289,6 +289,15 @@ Timeout can be passed to Devnet's HTTP server. This makes it easier to deploy an
 cargo run -- --timeout TIMEOUT
 ```
 
+### State archive mode
+
+With state archive capacity set to `full` devnet will store full state history. The default mode is `none`, where no old states are stored.
```
FabijanC, line 294: Needs the comma, and perhaps the project name can be capitalized:

```suggestion
With state archive capacity set to `full`, Devnet will store full state history. The default mode is `none`, where no old states are stored.
```

FabijanC (changes requested):

crates/starknet/src/starknet/get_class_impls.rs
mikiw: Fixed in c727b6cd25e1aaa1debbc6ded48ac22872545c48

crates/starknet/src/starknet/starknet_config.rs
mikiw: Fixed in c727b6cd25e1aaa1debbc6ded48ac22872545c48

README.md
mikiw: Fixed in c727b6cd25e1aaa1debbc6ded48ac22872545c48

FabijanC (approved):

PR #295, opened 2023-12-18
Add support of WEI/FRI minting

mikiw:
## Usage related changes

- Add WEI/FRI minting support

## Checklist:

- [x] Documented the changes
- [x] Updated the tests

### DIFF
--- a/crates/starknet-server/src/api/http/endpoints/mint_token.rs
+++ b/crates/starknet-server/src/api/http/endpoints/mint_token.rs
@@ -1,5 +1,5 @@
 use axum::{Extension, Json};
-use starknet_core::constants::ETH_ERC20_CONTRACT_ADDRESS;
+use starknet_core::constants::{ETH_ERC20_CONTRACT_ADDRESS, STRK_ERC20_CONTRACT_ADDRESS};
 use starknet_core::starknet::Starknet;
 use starknet_rs_core::types::{BlockId, BlockTag};
 use starknet_types::contract_address::ContractAddress;
@@ -16,8 +16,8 @@ pub(crate) async fn get_fee_token() -> HttpApiResult<Json<FeeToken>> {
 }
 
 /// get the balance of the `address`
-fn get_balance(starknet: &Starknet, address: ContractAddress) -> Result<BigUint, ApiError> {
-    let erc20_address = Felt::from_prefixed_hex_str(ETH_ERC20_CONTRACT_ADDRESS).unwrap();
+fn get_balance(starknet: &Starknet, address: ContractAddress, erc20_contract: &str) -> Result<BigUint, ApiError> {
+    let erc20_address = Felt::from_prefixed_hex_str(erc20_contract).unwrap();
     let balance_selector =
         starknet_rs_core::utils::get_selector_from_name("balanceOf").unwrap().into();
     let new_balance_raw = starknet.call(
@@ -46,20 +46,34 @@ pub(crate) async fn mint(
     Json(request): Json<MintTokensRequest>,
     Extension(state): Extension<HttpApiHandler>,
 ) -> HttpApiResult<Json<MintTokensResponse>> {
-    // increase balance
     let mut starknet = state.api.starknet.write().await;
+    let mut unit = "WEI".to_string();
+    let mut erc20_contract = ETH_ERC20_CONTRACT_ADDRESS;
 
+    // if unit is FRI, change contract address and unit
+    match request.unit {
+        Some(u) => {
+            if u == "FRI"
+            {
+                erc20_contract = STRK_ERC20_CONTRACT_ADDRESS;
+                unit = u
+            }
+        },
+        None => {},
+    }
+
+    // increase balance
     let tx_hash = starknet
-        .mint_wei(request.address, request.amount)
+        .mint(request.address, request.amount, erc20_contract)
         .await
         .map_err(|err| HttpApiError::MintingError { msg: err.to_string() })?;
 
-    let new_balance = get_balance(&starknet, request.address)
+    let new_balance = get_balance(&starknet, request.address, erc20_contract)
         .map_err(|err| HttpApiError::MintingError { msg: err.to_string() })?;
 
     Ok(Json(MintTokensResponse {
         new_balance: new_balance.to_str_radix(10),
-        unit: "WEI".to_string(),
-        tx_hash,
+        unit: unit,
+        tx_hash: tx_hash,
     }))
 }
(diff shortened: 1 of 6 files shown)


mikiw:
@FabijanC this WEI/FRI minting is just a temporary transition?

I'm not sure how to handle that in tests. Should I add a new test that will be removed in the future?

FabijanC:
> @mikiw:
> I'm not sure how to handle that in tests. Should I add a new test that will be removed in the future?

We at least need the following cases where we would assert response content:
- test which asserts `WEI` can be specified
- test which asserts `FRI` can be specified
- test which asserts that some random currency is rejected

Ideally we would also have this kind of test:
- predeploy accounts with balance 0
- try sending declare v3 tx - expect failure
- mint some STRK to the account (keeping WEI at 0)
- now tx succeeds

mikiw:
> > @mikiw:
> > I'm not sure how to handle that in tests. Should I add a new test that will be removed in the future?
> 
> My feeling is that we will need to support WEI for several more months.
> 
> We at least need the following cases where we would assert response content:
> 
> * test which asserts `WEI` can be specified
> * test which asserts `FRI` can be specified
> * test which asserts that not specifying a currency defaults to `WEI`
> * test which asserts that some random currency is rejected
> 
> Ideally we would also have this kind of test:
> 
> 1. predeploy accounts with balance 0
> 2. try sending declare v3 tx - expect failure
> 3. mint some STRK to the account (keeping WEI at 0)
> 4. now tx succeeds

Shouldn't this second scenario be tested in v3 tests?

FabijanC:
> @mikiw:
> Shouldn't this second scenario be tested in v3 tests?

If you don't feel like adding a manual request (i.e. if it's too complicated without starknet-rs), then just add a test which will be ignored (using `#[ignore = ...` as elsewhere in the code), but with `todo!()` in the test body (and perhaps with the steps from my previous post left as a comment).

mikiw:
> > @mikiw:
> > Shouldn't this second scenario be tested in v3 tests?
> 
> If you don't feel like adding a manual request (i.e. if it's too complicated without starknet-rs), then just add a test which will be ignored (using `#[ignore = ...` as elsewhere in the code), but with `todo!()` in the test body (and perhaps with the steps from my previous post left as a comment).

I would skip that v3 mint test, we also have `add_declare_v3_transaction_successful_execution ` and `declare_transaction_v3_with_max_fee_zero_should_return_an_error ` tests that cover this scenario separately

crates/starknet-server/src/api/http/endpoints/mint_token.rs
```diff
@@ -46,20 +50,30 @@ pub(crate) async fn mint(
...
+    if let Some(u) = request.unit {
+        if u == "FRI" {
+            erc20_contract = STRK_ERC20_CONTRACT_ADDRESS;
+            unit = u;
+        } else if u != "WEI" {
+            return Err(HttpApiError::MintingError {
+                msg: "Minting is supported only for WEI/FRI unit".to_string(),
+            });
+        }
+    }
```
FabijanC, line 67: A more idiomatic and concise way would be to use `unwrap_or` and `match`. An even better approach would be to use an enum, preferably an existing one, but here's how it should look without `if`s and `else`s. I also changed variable name from `erc20_contract` to `erc20_address`:

```suggestion
    let unit = request.unit.unwrap_or("WEI".into());

    let erc20_address = match unit.as_str() {
        "WEI" => ETH_ERC20_CONTRACT_ADDRESS,
        "FRI" => STRK_ERC20_CONTRACT_ADDRESS,
        other => {
            return Err(HttpApiError::MintingError {
                msg: format!("Minting is supported only for WEI/FRI, got: {other}"),
            })
        }
    };
```

crates/starknet/src/starknet/mod.rs
```diff
@@ -629,7 +629,12 @@ impl Starknet {
     }
 
     /// Creates an invoke tx for minting, using the chargeable account.
-    pub async fn mint(&mut self, address: ContractAddress, amount: u128) -> DevnetResult<Felt> {
+    pub async fn mint(
+        &mut self,
+        address: ContractAddress,
+        amount: u128,
+        erc20_contract: &str,
```
FabijanC, line 636: Can the name reflect the fact that it's an address abstraction, not a contract one?

FabijanC (changes requested):

crates/starknet-server/src/api/http/endpoints/mint_token.rs
```diff
@@ -16,8 +16,12 @@ pub(crate) async fn get_fee_token() -> HttpApiResult<Json<FeeToken>> {
...
 
 /// get the balance of the `address`
-fn get_balance(starknet: &Starknet, address: ContractAddress) -> Result<BigUint, ApiError> {
-    let erc20_address = Felt::from_prefixed_hex_str(ETH_ERC20_CONTRACT_ADDRESS).unwrap();
+fn get_balance(
+    starknet: &Starknet,
+    address: ContractAddress,
+    erc20_contract: &str,
+) -> Result<BigUint, ApiError> {
+    let erc20_address = Felt::from_prefixed_hex_str(erc20_contract).unwrap();
```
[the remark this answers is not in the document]
FabijanC: I completely missed this unwrap. Perhaps the argument of this function should not be a `&str`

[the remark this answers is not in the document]
mikiw: true, true I'll change that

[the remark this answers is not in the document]
mikiw: fixed in 456aa68b7662dbb9fc6374ff109ab36e2943be97

crates/starknet-server/src/api/http/endpoints/mint_token.rs
```diff
@@ -46,20 +50,27 @@ pub(crate) async fn mint(
...
+        Some(FeeUnits::WEI) | None => {
+            ContractAddress::new(Felt::from_prefixed_hex_str(ETH_ERC20_CONTRACT_ADDRESS).unwrap())
+                .unwrap()
+        }
+        Some(FeeUnits::FRI) => {
+            ContractAddress::new(Felt::from_prefixed_hex_str(STRK_ERC20_CONTRACT_ADDRESS).unwrap())
+                .unwrap()
+        }
+    };
+    let unit = if let Some(unit) = request.unit { unit } else { FeeUnits::WEI };
```
FabijanC, line 64: This unnecessary logic is introduced again.
This should be done using `.unwrap_or` before `match`, and then apply `match` on the unwrapped object.

FabijanC (changes requested):

crates/starknet-server/src/api/http/endpoints/mint_token.rs
mikiw: Right somehow I missed that, mainly because I didn't know that it would behave like this in case of `None` 

Fixed in e1a993eb0b1dcdd95373ec7a776ac8242b952388

FabijanC (approved):

PR #296, opened 2023-12-18
fix: ensure `from_address` is correctly set from blockifier

glihm:
## Usage related changes

Fixes some situations of messaging when the message is sent by a contract call, and not a library call - relies on storage address.

When using the script `run_e2e.sh`, it's important to use scarb `2.3.1` for now as starkli is not release with `2.4.0` compiler version.

I'm sorry for missing that earlier, it was found debugging and comparing messaging with Devnet, Katana (were we also had this issue) and Testnet.

## Checklist:

- [X] Applied formatting - `./scripts/format.sh`
- [X] No linter errors - `./scripts/clippy_check.sh`
- [X] No unused dependencies - `./scripts/check_unused_deps.sh`
- [X] Performed code self-review
- [X] Rebased to the last commit of the target branch (or merged it into my branch)
- [X] Documented the changes
- [X] Linked the issues which this PR resolves
- [X] Checked the TODO section in README.md if this PR resolves it
- [X] Updated the tests
- [X] All tests are passing - `cargo test`


crates/starknet/src/transactions.rs
```diff
@@ -219,14 +219,22 @@ impl StarknetTransaction {
...
+            };
 
             messages.extend(call_info.execution.l2_to_l1_messages.iter().map(|m| {
                 OrderedMessageToL1 {
                     order: m.order,
                     message: MessageToL1 {
                         to_address: m.message.to_address.into(),
-                        from_address,
+                        from_address: ContractAddress::new(from_address.into())
+                            .expect("Invalid contract address from blockifier's messsage"),
```
FabijanC, line 237: @marioiordanov Do you have a better idea about this? Seems like changing `get_blockifier_messages_recursively` to return `Result` will only cause more changes

FabijanC (changes requested):

crates/starknet/src/transactions.rs
marioiordanov: right, it is used in two places, which will require cascade style changes. But I  dont see the method signature being changed

marioiordanov, line 237: Lets use the error propagation to avoid any panics in the code.

FabijanC: I didn't mean that the code is already changing the method. I meant that the usual way of fixing this is by propagating the error, which would cause changing the type to `Result` and defining the needed errors (and first refactoring `.iter().map(...)` to a for-loop.

marioiordanov: yes, right

FabijanC:
Related: https://github.com/0xSpaceShard/starknet-devnet-rs/pull/298

crates/starknet/src/transactions.rs
FabijanC: It seems that if we use
```rust
            let from_address = if let Some(code_address) = call_info.call.code_address {
                code_address
            } else {
                call_info.call.caller_address
            };
```
then we can simply do
```rust
                    message: MessageToL1 {
                        ...,
                        from_address: from_address.into(),
                        ...
                    }
```

But I'm still taking a look at #298 which is addressing the same part of the code

PR #297, opened 2023-12-19
Integration test error fix

marioiordanov:

- Fixes CI fail after last merge to main
- adds data to ValidationFailure error, because starknet-rs fails to deserialise if no "data" is provided to the JSON object if the error is with code 55 (ValidationFailure)

## Checklist:

- [ ] Documented the changes
- [ ] Checked the TODO section in README.md if this PR resolves it
- [ ] Updated the tests

### DIFF
--- a/crates/starknet-server/src/api/json_rpc/error.rs
+++ b/crates/starknet-server/src/api/json_rpc/error.rs
@@ -25,7 +25,7 @@ pub enum ApiError {
     InvalidTransactionIndexInBlock,
     #[error("Class hash not found")]
     ClassHashNotFound,
-    #[error("Contract error: {error}")]
+    #[error("Contract error")]
     ContractError { error: starknet_core::error::Error },
     #[error("There are no blocks")]
     NoBlocks,
@@ -49,8 +49,8 @@ pub enum ApiError {
     InsufficientMaxFee,
     #[error("Account balance is smaller than the transaction's max_fee")]
     InsufficientAccountBalance,
-    #[error("Account validation failed: {reason}")]
-    ValidationFailure { reason: String, data: String },
+    #[error("Account validation failed")]
+    ValidationFailure { reason: String },
     #[error("No trace available for transaction")]
     NoTraceAvailable,
 }
@@ -154,10 +154,10 @@ impl ApiError {
                 message: error_message.into(),
                 data: None,
             },
-            ApiError::ValidationFailure { data, .. } => RpcError {
+            ApiError::ValidationFailure { reason } => RpcError {
                 code: server::rpc_core::error::ErrorCode::ServerError(55),
                 message: error_message.into(),
-                data: Some(serde_json::Value::String(data)),
+                data: Some(serde_json::Value::String(reason)),
             },
             ApiError::StarknetDevnetError(
                 starknet_core::error::Error::TransactionValidationError(validation_error),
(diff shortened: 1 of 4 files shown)


mikiw (approved):

crates/starknet-server/src/api/json_rpc/error.rs
```diff
@@ -50,7 +50,7 @@ pub enum ApiError {
     #[error("Account balance is smaller than the transaction's max_fee")]
     InsufficientAccountBalance,
     #[error("Account validation failed: {reason}")]
-    ValidationFailure { reason: String },
+    ValidationFailure { reason: String, data: String },
```
FabijanC, line 53: Shouldn't we just rename `reason` to `data`? This way it looks like we have two overlapping properties.

Looking at [the spec](https://github.com/starkware-libs/starknet-specs/blob/v0.6.0/api/starknet_write_api.json#L262), I see it was my mistake to use the name `reason` in #287

FabijanC (changes requested):

crates/starknet-server/src/api/json_rpc/error.rs
FabijanC: Actually the name reason is not so much of a problem, but I misused the `message` property

marioiordanov: applied a fix [a36b0d6](https://github.com/0xSpaceShard/starknet-devnet-rs/pull/297/commits/a36b0d653423b3fffced06474bad22e6350b2184)

FabijanC: I am confused now, didn't this completely remove the `data` property, which is required by the spec (and by starknet-rs)?

FabijanC: Ah ok, it does not, it's defined in RpcError

marioiordanov: no, because now the `data` property is populated from the `reason`

crates/starknet-server/src/api/json_rpc/error.rs
```diff
@@ -25,7 +25,7 @@ pub enum ApiError {
     InvalidTransactionIndexInBlock,
     #[error("Class hash not found")]
     ClassHashNotFound,
-    #[error("Contract error: {error}")]
+    #[error("Contract error")]
```
FabijanC, line 28: According to the specification, shouldn't there also be some `data` in this error: https://github.com/starkware-libs/starknet-specs/blob/49665932a97f8fdef7ac5869755d2858c5e3a687/api/starknet_api_openrpc.json#L3903-L3916

Otherwise we are again hiding the real cause of error and making it hard for users to debug. Maybe this can be handled in a separate PR?

crates/starknet-server/src/api/json_rpc/error.rs
FabijanC: Yes, I realized that between my last two comments. I'd missed the `data: ` part in `RpcError` definition

crates/starknet-server/src/api/json_rpc/error.rs
marioiordanov: the users should use the data property because the spec defines only "Contract error" as data to the `message` field

FabijanC: Mistake on my end, didn't see where ContractError is converted to RpcError. Resolving.

FabijanC (approved):

PR #315, opened 2024-01-18
Add 0.8.0 OZ contracts

mikiw:
## Usage related changes

- support of 0.8.0 version of OZ account contract and ERC20 contract was added, users can still use 0.5.1 with `--account-class cario0`, internal chargeable account is still using 0.5.1
- cairo 1 accounts are now predeployed the default (used to be cairo 0)

- mint function implementation is missing from ERC20_Mintable_OZ_0.8.0.json, OpenZeppelin is providing [wizard](https://wizard.openzeppelin.com/cairo) with mintable feature implementation that, we used.

## Checklist:

- [x] Documented the changes
- [x] Updated the tests


mikiw:
I'm still on the 0.8.1 compilation error topic, I might update this PR to be 0.8.1 not 0.8.0

mikiw:
PR is ready for review again. I needed to fix one test after the rebase.

And about 0.8.1 - it can be handled in new PR.

crates/starknet-devnet-core/src/starknet/predeployed.rs
```diff
@@ -4,26 +4,23 @@ use starknet_types::contract_storage_key::ContractStorageKey;
...
-        })?;
-    let erc20_fee_contract = SystemContract::new(
-        ERC20_CONTRACT_CLASS_HASH,
-        contract_address,
-        &erc20_contract_class_json_str,
-    )?;
+pub(crate) fn create_erc20_at_address(
+    contract_address: &str,
+    path: &str,
+    hash: &str,
```
FabijanC, line 18: Why should this now be specified if in production we will always use the same cairo 1 implementation? Shouldn't the path and the hash be hardcoded?

FabijanC (changes requested):

crates/starknet-devnet-core/src/starknet/predeployed.rs
mikiw: Thanks for the review. It's a remnant of my refactor where I was mixing both versions.

Fixing...

mikiw: Fixed in 94bd3f331a5ae81dadf01b0cf9efe8a451c29bfd

FabijanC (approved):

Issue #195, opened 2023-10-06
Forking


FabijanC:
We can only implement it as it was in Devnet-py if the underlying Starknet state implementation has a separate State and StateReader.

- [x] First we need to switch to blockifier.

maciejka:
I will give it a try if possible.

PR #329, opened 2024-01-31
Pre-forking state refactor

### LINKED ISSUE #308: Recompilation from sierra to casm is done each time casm is fetched
**Describe the bug (observed vs expected behavior)**

- This is not really a bug because everything is functionally as it should be, but there is unnecessary overhead.
- Casm should be generated from sierra only once (and stored), then retrieved when needed, instead of storing only the sierra and recompiling on each invoke/call/deploy of a Cairo 1 contract.

**To Reproduce**

1. Put a debug breakpoint or print line in Cairo1 case of this conversion code https://github.com/0xSpaceShard/starknet-devnet-rs/blob/5d2536a99852b1a61bbbfdcaa6755cb4275bffdd/crates/types/src/rpc/contract_class.rs#L129-L132
2.  Run devnet with `cargo run -- --seed 42 --account-class cairo1`
    - reproducible with any Cairo 1 contract, but for simplicity relying on predeployed account
3. Run the following (getting public key of first predeployed account):
```bash
curl localhost:5050 --json '{
	"jsonrpc": "2.0",
	"method": "starknet_call",
	"params": {
        "request": {
            "contract_address": "0x34ba56f92265f0868c57d3fe72ecab144fc96f97954bbbc4252cef8e8a979ba",
            "entry_point_selector": "0x01a6c6a0bdec86cc645c91997d8eea83e87148659e3e61122f72361fd5e94079",
            "calldata": []
        },
        "block_id": "latest"
    },
    "id": 0
}'
```

FabijanC:
## Usage related changes

- Needed for #195
- ~Temporarily disable `--state-archive-capacity full` (shall be restored by #353)~
- Improve performance: close #308

- Ignore tests that rely on full state archive (9 tests)
- Unify declaring and storing of class in `StarknetState`
- Unify state diff calculation and state clearing
  - Dirty vs clean state terminology replaced by relying on `CachedState` commitment diff
  - No more explicit applying of state diff
  - Class diff done via new utility struct `CommittedClassStorage` (property of `StarknetState`)
- Introduce `DictState`
  - Copy of blockifier's `DictStateReader` testing utility with added `State` (writing) functionality
  - Used for predeclaration and predeployment
    - Prevents invalid state diff (since state diff is now calculated using CachedState diff
  - Currently `CachedState` always takes `DictState` as its underlying state, but in a future PR this will be made generic to support a `ForkedState` that will be the same as `DictState`, but will also an origin
- Setting initial balance is now a part of account predeployment
- Remove `StateChanger`: where applicable, replaced with `blockifier::State`
- Remove `StateExtractor`: where applicable, replaced with `blockifier::StateReader`
- Introduce `CustomState` and `CustomStateReader`
  - Seems redundant as no function parameters use `impl CustomStateReader`, the methods of these traits could be implemented directly under `StarknetState`
- Since we are more relying on CachedState now and many of its methods require `&mut state`
  - This mutability requirement propagates to other parts of our code
  - Lock acquiring changed in some places from `state.read()` to `state.write()` accordingly
- Proposals regarding things not changed by this PR (perhaps these need separate GitHub issues):
  - Rename trait `Deployed` to `Predeployed` and its methods accordingly (predeploy, predeclare)
  - Block context needs to be stored together with state for correct old state calls
  - Do something about `StateDiff`:
    - In production it is only used to be converted to `ThinStateDiff`
    - Other use is only in testing

## Checklist:

- [x] Documented the changes
- [x] Updated the tests


FabijanC:
While working on #360, I spotted a bug in state diff generation. Converting to draft until fixed.

crates/starknet-devnet-server/src/api/json_rpc/mod.rs
```diff
@@ -83,15 +83,15 @@ pub struct JsonRpcHandler {
 impl RpcHandler for JsonRpcHandler {
     type Request = StarknetRequest;
 
-    async fn on_request(&self, request: Self::Request) -> ResponseResult {
+    async fn on_request(&mut self, request: Self::Request) -> ResponseResult {
```
marioiordanov, line 86: why the reference is being changed to mut

crates/starknet-devnet-core/src/state/state_readers.rs
```diff
@@ -0,0 +1,121 @@
...
+use blockifier::state::errors::StateError;
+use blockifier::state::state_api::{State, StateReader, StateResult};
+use starknet_api::core::{ClassHash, CompiledClassHash, ContractAddress, Nonce};
+use starknet_api::hash::StarkFelt;
+use starknet_api::state::StorageKey;
+
+/// A simple implementation of `StateReader` using `HashMap`s as storage.
+/// Copied from blockifier test_utils, added `impl State`
+#[derive(Debug, Default, Clone)]
+pub struct DictState {
```
marioiordanov, line 14: is there a difference between DictState and DictStateReader? If no why not use it as an internal implementation of DictState. Like this: 
```rust
pub struct DictState(pub(crate) DictStateReader)
```

FabijanC: The problem is in the ability to clone the struct. If I used `DictStateReader` as an internal struct, I'd still have to write my own `impl Clone`:
```rust
/// A simple implementation of `StateReader` using `HashMap`s as storage.
/// Copied from blockifier test_utils, added `impl State`
#[derive(Debug, Default)]
pub struct DictState(pub (crate) DictStateReader);

impl Clone for DictState {
    fn clone(&self) -> Self {
        Self(DictStateReader {
            storage_view: self.0.storage_view.clone(),
            address_to_nonce: self.0.address_to_nonce.clone(),
            address_to_class_hash: self.0.address_to_class_hash.clone(),
            class_hash_to_class: self.0.class_hash_to_class.clone(),
            class_hash_to_compiled_class_hash: self.0.class_hash_to_compiled_class_hash.clone(),
        })
    }
}
```

If we go this way, would we still have to provide the implementation of each `StateReader` method? By replacing e.g. `self.storage_view` with `self.0.storage_view`? Or would there be something smarter?

FabijanC: By switching `pub struct DictState(pub(crate) DictStateReader);`, we would use less imports, but would need four more code lines. Not sure which way to go

crates/starknet-devnet-server/src/api/json_rpc/mod.rs
FabijanC: I was sure that it didn't compile without `mut`; looks like it's compiling. Addressed in https://github.com/0xSpaceShard/starknet-devnet-rs/pull/329/commits/1999c72cc0f02d2e098c826899446d00952f11ab

crates/starknet-devnet-core/src/starknet/get_class_impls.rs
```diff
@@ -1,38 +1,44 @@
...
 pub fn get_class_hash_at_impl(
-    starknet: &Starknet,
+    starknet: &mut Starknet,
     block_id: &BlockId,
     contract_address: ContractAddress,
 ) -> DevnetResult<ClassHash> {
-    let state = starknet.get_state_at(block_id)?;
-    let class_hash = state.state.state.class_hash_at(&contract_address);
+    let state = starknet.get_mut_state_at(block_id)?;
+    state.assert_contract_deployed(contract_address)?;
```
marioiordanov, line 17: assert_contract_deployed could be omitted, because internally it uses is_contract_deployed which uses get_class_hash_at

crates/starknet-devnet-core/src/state/mod.rs
```diff
@@ -3,23 +3,79 @@ use std::collections::HashMap;
...
-use crate::error::{DevnetResult, Error, StateError};
-use crate::traits::{DevnetStateReader, StateChanger, StateExtractor};
+use self::state_readers::DictState;
+use crate::error::{DevnetResult, Error};
 
 pub(crate) mod state_diff;
+mod state_readers;
 pub mod state_update;
 
+pub trait CustomStateReader {
```
marioiordanov, line 22: can those traits be moved to traits.rs

crates/starknet-devnet-core/src/state/state_readers.rs
marioiordanov: I see that both approaches have their tradeoffs. I will resolve this conversation

crates/starknet-devnet-core/src/state/state_readers.rs
```diff
@@ -0,0 +1,121 @@
...
+        self.class_hash_to_compiled_class_hash.insert(class_hash, compiled_class_hash);
+        Ok(())
+    }
+
+    fn to_state_diff(&mut self) -> blockifier::state::cached_state::CommitmentStateDiff {
+        panic!("Should never be called")
+    }
+
+    fn add_visited_pcs(&mut self, _class_hash: ClassHash, _pcs: &std::collections::HashSet<usize>) {
+        todo!("What with this")
```
FabijanC, line 119: @marioiordanov I was thinking of simply adding a `panic!` here, similar to the method above

crates/starknet-devnet-core/src/state/mod.rs
FabijanC: They could, but honestly, having a dedicated traits.rs file for storing all traits looks like an antipattern to me. Maybe it would be more meaningful if it was in `state_readers.rs`.

crates/starknet-devnet-core/src/state/mod.rs
```diff
@@ -113,538 +244,253 @@ impl blockifier::state::state_api::StateReader for DevnetState {
...
-    /// this method clears the state from data that was accumulated in the StateCache
-    /// and restores it to the data in the state_reader, which is the "persistent" data
-    pub(crate) fn clear_dirty_state(&mut self) {
-        self.state = CachedState::new(
-            self.state.state.clone(),
-            GlobalContractCache::new(GLOBAL_CONTRACT_CACHE_SIZE_FOR_TEST),
-        );
+impl CustomStateReader for StarknetState {
+    fn is_contract_deployed(&mut self, contract_address: ContractAddress) -> bool {
+        let api_address = contract_address.try_into().unwrap();
```
FabijanC, line 260: While looking into the contract deployment assertion, I realized there's this `try_into().unwrap()` which I had left to think about later. Is that an acceptable one?

crates/starknet-devnet-core/src/starknet/get_class_impls.rs
marioiordanov: `assert_contract_deployed` uses self.get_class_hash_at and checks if its not StarkFelt::Zero. After the method invokation the same check is done again

crates/starknet-devnet-core/src/state/mod.rs
marioiordanov: I missed it, but not acceptable

crates/starknet-devnet-core/src/starknet/get_class_impls.rs
FabijanC: Ah, alright, you mean omitting the assertion just in this place. I thought you meant generally dropping the `assert_contract_deployed` logic. Will address this.

FabijanC: Removed in https://github.com/0xSpaceShard/starknet-devnet-rs/pull/329/commits/0083e8f90b1370896120c3aa5edfda535b256a77

crates/starknet-devnet-core/src/state/mod.rs
FabijanC: Changed in https://github.com/0xSpaceShard/starknet-devnet-rs/pull/329/commits/ac7329ee50833d18eaf23ff50ee8061a648bc0ff

crates/starknet-devnet-core/src/state/state_readers.rs
marioiordanov: I think a better one is to be with empty implementation

FabijanC: Warning added in https://github.com/0xSpaceShard/starknet-devnet-rs/pull/329/commits/ee01d66cf3ab6fd8d3fab36120c34c7f89af5037

FabijanC: Redundant methods removed in commits:
- https://github.com/0xSpaceShard/starknet-devnet-rs/pull/329/commits/6696dc6ef1a78deb34b0ac62a3e1250d926f437f
- https://github.com/0xSpaceShard/starknet-devnet-rs/pull/329/commits/41fa1979abe3489c269b24fd46c705c9d24b6a7b

marioiordanov (approved):

PR #338, opened 2024-02-07
Use more idiomatic bash

FabijanC:
## Usage related changes

- None

- Use more idiomatic bash approach
- Reduce script complexity
- Apply formatting in a couple of files
  - using VS Code extensions: `Bash IDE` and `Even Better TOML`
- Machete: ignore unused cairo-lang dependencies

## Checklist:

- [x] Documented the changes
- [ ] Updated the tests

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -97,7 +97,7 @@ jobs:
             ARCH_SUFFIX: -arm
             REMOTE: ""
 
-  tag-and-create-joint-image-manifest:
+  publish:
     docker:
       - image: cimg/rust:1.70.0
     resource_class: medium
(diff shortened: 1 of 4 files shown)


marioiordanov (commented):
how the quotes solve the problem for missing argument in the create_and_push_manifest function, when passed "" as seed_prefix

FabijanC:
> how the quotes solve the problem for missing argument in the create_and_push_manifest function, when passed "" as seed_prefix

That's how argument parsing works in bash, I guess. Here's a minimum reproducible example:
```bash
set -eu

function foo() {
    echo $1
}

var=""
foo "$var"
```
Try removing the quotes in `foo "$var"` and see how it fails

marioiordanov (approved):

PR #356, opened 2024-02-21
Comment out of owner and spender [key] attributes in erc20 contract

mikiw:
## Usage related changes

- in erc20 contract in struct Approval, owner and spender [key] attributes were removed similar to
https://github.com/starknet-io/starkgate-contracts/blob/d62a255307d2f3de65665f18316766a2c69ead78/src/openzeppelin/token/erc20_v070/erc20.cairo#L100
- erc20 class hash was hardcoded to be the same as OZ class hash which is https://docs.openzeppelin.com/contracts-cairo/0.8.1/presets although it should be different due to commented key attributes and add of mintable feature

## Checklist:

- [x] Documented the changes
- [x] Updated the tests


crates/starknet-devnet-core/src/constants.rs
```diff
@@ -23,7 +23,7 @@ pub const CAIRO_1_ERC20_CONTRACT_PATH: &str =
     concat!(env!("CARGO_MANIFEST_DIR"), "/accounts_artifacts/ERC20_Mintable_OZ_0.8.1.json");
 
 pub const CAIRO_1_ERC20_CONTRACT_CLASS_HASH: &str =
-    "0x02ce20303e4b963bed8f11fdb309f859822ee17be99399a9acdc289443bc9d09";
+    "0x046ded64ae2dead6448e247234bab192a9c483644395b66f2155f2614e5804b0";
```
FabijanC, line 26: Add a comment explaining why this is hardcoded, possibly providing the OZ link

FabijanC (changes requested):

crates/starknet-devnet-core/src/constants.rs
mikiw: Added in ff6daa2769de7b9a4b1bc32a2fcac4676083ca4c

FabijanC: Please don't kill me, but missing one slash to make it triple instead of double in L27.

PR #377, opened 2024-03-12
Fix account balance overflow

### LINKED ISSUE #110: Trait Accounted treating balance in a wrong way
The balance is stored in two memory places (two consecutive storage addresses) whereas we only deal with one (both in `set_initial_balance` and `get_balance`). This lowers the upper bound on the values we can store. E.g. in Pythonic Devnet the chargeable account was using an initial balance of 2^251 so I wanted to set it here as well, but 2^128 is the maximum. We can probably live with 2^128 for the chargeable account, but we will have problems if a user specifies account initial balance to be greater than that.

In Pythonic Devnet, the balance is set like this:
https://github.com/0xSpaceShard/starknet-devnet/blob/7840c9dc1735d9bebb9d7bd0264714a502587c0b/starknet_devnet/account_util.py#L119-L140

JimmyFate:
Closes #110 

## Usage related changes

- ~~amount must be a string instead of plain integer in `/mint`~~

- [add primitive-types](https://github.com/0xSpaceShard/starknet-devnet-rs/commit/66f5c8014f919ac8c901b1e2e83003a7a9ccf737)
- [change balance to u256](https://github.com/0xSpaceShard/starknet-devnet-rs/commit/f484e434f1c68c669d27c2c7a20d0647067e5520)

## Checklist:

- [ ] Documented the changes
- [x] Updated the tests

---

These 2 tests are failing locally for me:

```sh
failures:
    test_messaging::can_deploy_l1_messaging_contract
    test_messaging::can_interact_with_l1
```


FabijanC (changes requested):
1. Why do we need a new dependency for all this (`primitive_types::U256`)? Can't we use the existing `BigUint`?
2. I don't know if we want to introduce the breaking semantic change of requiring a string instead of a number in the /mint endpoint. Can't we deserialize a big numeric JSON literal into the desired numeric Rust type (be it U256 or BigUint), without it needing to be a string?

cc: @marioiordanov

JimmyFate:
> 1. Why do we need a new dependency for all this (`primitive_types::U256`)? Can't we use the existing `BigUint`?

It'd require creating a custom `U256` struct wrapping the `BigUint`, with other impls, bloating the codebase. I feel like it's better to just use an existing one.

> 2. I don't know if we want to introduce the breaking semantic change of requiring a string instead of a number in the /mint endpoint. Can't we deserialize a big numeric JSON literal into the desired numeric Rust type (be it U256 or BigUint), without it needing to be a string?

Rust's built-in numeric types only go up to u128, so there's no built-in way to deserialize a u256 directly. I've you have any idea on how to do this, I'm down to change it.

crates/starknet-devnet-core/src/account.rs
```diff
@@ -103,12 +104,14 @@ impl Account {
         Ok(ContractAddress::from(account_address))
     }
 
+    #[allow(dead_code)]
```
marioiordanov, line 107: If those functions are marked with 'dead_code' why not remove them

marioiordanov:
> > 1. Why do we need a new dependency for all this (`primitive_types::U256`)? Can't we use the existing `BigUint`?
> 
> It'd require creating a custom `U256` struct wrapping the `BigUint`, with other impls, bloating the codebase. I feel like it's better to just use an existing one.
> 
> > 2. I don't know if we want to introduce the breaking semantic change of requiring a string instead of a number in the /mint endpoint. Can't we deserialize a big numeric JSON literal into the desired numeric Rust type (be it U256 or BigUint), without it needing to be a string?
> 
> Rust's built-in numeric types only go up to u128, so there's no built-in way to deserialize a u256 directly. I've you have any idea on how to do this, I'm down to change it.

It's possible for deserialising directly from some number. Just deserialize to serde_json::Value::Number

FabijanC:
> It's possible for deserialising directly from some number. Just deserialize to serde_json::Value::Number. Then convert to string and then convert to U256

I ran an experiment and it seems that precision is lost that way: [experiment link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=209a2e26733e62781a1f2f390da0ca1a)

marioiordanov:
> experiment link

The experiment doesn't emulate starknet-devnet-rs, because it uses arbitrary precision in the crate that the file is

FabijanC:
> > experiment link
> 
> The experiment doesn't emulate starknet-devnet-rs, because it uses arbitrary precision in the crate that the file is

Wow, didn't know that. I ran the code as a devnet test and indeede the precision is preserved.

JimmyFate:
> It's possible for deserialising directly from some number. Just deserialize to serde_json::Value::Number. Then convert to string and then convert to U256

Thanks for the great tip @marioiordanov! Didn't know that.
Done as pointed out 👍

FabijanC (commented):
Can we now lose the U256 dependency?

JimmyFate:
> Can we now lose the U256 dependency?

Could `ethers` be used instead or do you insist on a separate own struct for U256? `starknet-devnet-core` already uses it (it uses `primitive-types` itself too)

marioiordanov (approved):

FabijanC (approved):
> > Can we now lose the U256 dependency?
> 
> Could `ethers` be used instead or do you insist on a separate own struct for U256? `starknet-devnet-core` already uses it (it uses `primitive-types` itself too)

I meant losing the `primitive-types` dependency. But I guess an extra dependency is not such a big issue; your PR is good, so let's merge it. Thank you for the contribution and the patience.

PR #365, opened 2024-02-29
Introduce forking

FabijanC:
## Usage related changes

- Close #195
- User instructions are in `README.md` under `Forking`
  - basically it's a subset of [Devnet-py's](https://0xspaceshard.github.io/starknet-devnet/docs/guide/fork/) CLI params, but the origin interaciton possibilities should be the same (if not better)
- In case of `NoStateAtBlock`, a wildcard error is returned instead of `ContractError`

## Development related changes

- Add two forking structures: `OriginForwarder` and `StarknetDefaulter` (read more about them in the code docs)
- Use 6 parallel jobs in testing (previously had 7 - failing)
  - Related: #378
  - If #391 is merged before this PR, the conflict shall be resolved in favor of that PR
- Improve testing by adding utilities
  - e.g. `BackgroundDevnet` methods like `create_block`, `fork`, `get_balance_at_block`
- Two e2e tests interact with Sepolia.
- Replace `StateHistoryDisabled` error with `NoStateAtBlock`
- Add `get_block_number` method to `timestamp.cairo` testing contract and rename it to `block_reader.cairo`
- Refactor `crates/starknet-devnet-server/src/api/json_rpc/mod.rs` to call `.to_rpc_result()` in only one place.

## Checklist:

- [x] Applied formatting - `./scripts/format.sh`
- [x] No linter errors - `./scripts/clippy_check.sh`
- [x] No unused dependencies - `./scripts/check_unused_deps.sh`
- [x] Performed code self-review
- [x] Rebased to the last commit of the target branch (or merged it into my branch)
- [x] Documented the changes
- [x] Linked the issues which this PR resolves
- [x] Checked the TODO section in README.md if this PR resolves it
- [x] Updated the tests
- [x] All tests are passing - `cargo test`

PR #379, opened 2024-03-13
Forking example using starknet-rs

marioiordanov:

Added example for interacting with starknet-rs from sync code 

## Checklist:

- [ ] Documented the changes
- [ ] Checked the TODO section in README.md if this PR resolves it
- [ ] Updated the tests


crates/starknet-devnet-core/src/starknet/forked_provider.rs
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    async fn handle_message(msg: ForkedRequest, client: Arc<JsonRpcClient<HttpTransport>>) {
+        match msg {
+            ForkedRequest::GetNonceAt(address, sender) => {
+                let nonce = client
+                    .get_nonce(BlockId::Tag(starknet_rs_core::types::BlockTag::Latest), address)
+                    .await;
+
+                sender.send(nonce).expect("unable to send result from get_nonce");
```
FabijanC, line 43: What's with this `expect`?

crates/starknet-devnet-core/src/starknet/forked_provider.rs
```diff
@@ -0,0 +1,81 @@
...
+        let (sender, receiver) = tokio::sync::mpsc::channel(10);
+        let provider = ForkedProvider::new(rpc_url, receiver);
+
+        std::thread::spawn(move || {
+            tokio::runtime::Builder::new_current_thread()
+                .enable_all()
+                .build()
+                .expect("failed to create tokio runtime")
+                .block_on(provider.run());
+        });
```
FabijanC, line 65: Are we assuming that this thread will be spawned before the incoming of the first request that would need it?

crates/starknet-devnet-core/src/starknet/forked_provider.rs
```diff
@@ -0,0 +1,108 @@
...
+        let json_rpc_client = JsonRpcClient::new(HttpTransport::new(fork_url));
+
+        Self { client: Arc::new(json_rpc_client), receiver: rx }
+    }
+
+    async fn run(mut self) {
+        while let Some(msg) = self.receiver.recv().await {
+            tokio::task::spawn(Self::handle_message(msg, self.client.clone()));
+        }
+    }
```
FabijanC, line 34: What if there are multiple incoming get_nonce requests? Then they will all push into the same channel? Is that a problem when receiving from the channel?

crates/starknet-devnet-core/src/starknet/defaulter.rs
```diff
@@ -64,6 +65,7 @@ impl BlockingOriginReader {
 #[derive(Clone, Debug, Default)]
 pub struct Defaulter {
     origin_reader: Option<BlockingOriginReader>,
+    forked_reader: Option<ForkedOrigin>,
```
FabijanC, line 68: Do we need both? Or did you just decide not to delete the one I added?

crates/starknet-devnet-core/src/starknet/forked_provider.rs
```diff
@@ -0,0 +1,108 @@
...
+                .build()
+                .expect("failed to create tokio runtime")
+                .block_on(provider.run());
+        });
+
+        Self { sender }
+    }
+
+    pub(crate) fn get_nonce(&self, address: FieldElement) -> DevnetResult<FieldElement> {
+        let (tx, rx) = mpsc::channel();
```
FabijanC, line 71: Why are we here creating another pair of tx+rx?

FabijanC (changes requested):

crates/starknet-devnet-core/src/starknet/forked_provider.rs
marioiordanov: to communicate between the two threads. The thread that get_nonce is running and the thread that ForkedProvider operates in

crates/starknet-devnet-core/src/starknet/forked_provider.rs
marioiordanov: It will be spawned most probably. All of the operations after creating the StarknetState are not atomic, its very little chance that the thread will not be spawned

FabijanC: Yes, sure, was just checking.

crates/starknet-devnet-core/src/starknet/forked_provider.rs
marioiordanov: The two channels send different data, one is for asking the new thread to start an async request, the other is for the async request to return data to the first thread.

FabijanC: Right, they are one-way channels. Resolving.

crates/starknet-devnet-core/src/starknet/forked_provider.rs
marioiordanov: This panics if the receiver end is disconnected. This means that the receiving channel crated in `ForkedOrigin::get_nonce` is dropped or went out of scope, therefore there is a bug. The other way is to ignore the result of sender.send. Error propagation is not possible, because there is no error to be returned from t `handle_message`

FabijanC: Ok, I don't think we should ignore. I'm ok with the panic if it indicates a bug, just thought that maybe you forgot to change it.

FabijanC:
Closed in favor of the `reqwest`-based approach in #365. If we decide that we want to take this channel-based approach, we can reopen.

PR #410, opened 2024-04-03
Make request body size limit configurable

### LINKED ISSUE #408: Request body size limit should be configurable
Either with a CLI flag (limit on/off), or a CLI param that accepts a size. Personally I prefer avoiding booleans/flags, so inclined to the latter approach.

Currently big requests are failing, so marking this as a bug.

In the flag approach, can be as simple as adding `.layer(DefaultBodyLimit::disable())` to [crates/starknet-devnet-server/src/builder.rs](https://github.com/0xSpaceShard/starknet-devnet-rs/blob/a147b4cd72f9ce9d1fa665d871231370db0f51c7/crates/starknet-devnet-server/src/builder.rs#L102)

FabijanC:
## Usage related changes

- Close #408
- Introduce a new CLI parameter: `--request-body-size-limit <BYTES>`
  - Default: `2000000` (2 MB) - so far this was the only value used
  - This is more flexible than implementing a boolean flag parameter

- Refactor: extract host, port, timeout and request_body_size_limit to ServerConfig
- Remove redundant code from ServerConfig

## Checklist:

- [x] Checked the relevant parts of [development docs](https://github.com/0xSpaceShard/starknet-devnet-rs/?tab=readme-ov-file#development)
- [x] Updated the docs if needed, including the [TODO section](https://github.com/0xSpaceShard/starknet-devnet-rs/?tab=readme-ov-file#todo-to-reach-feature-parity-with-the-pythonic-devnet)
- [x] Updated the tests if needed; all passing ([execution info](https://github.com/0xSpaceShard/starknet-devnet-rs/?tab=readme-ov-file#test-execution))

### DIFF
--- a/crates/starknet-devnet-server/src/builder.rs
+++ b/crates/starknet-devnet-server/src/builder.rs
@@ -2,6 +2,7 @@ use std::convert::Infallible;
 use std::net::SocketAddr;
 use std::time::Duration;
 
+use axum::extract::DefaultBodyLimit;
 use axum::response::Response;
 use axum::routing::{post, IntoMakeService};
 use axum::{Extension, Router};
@@ -95,11 +96,13 @@ impl<TJsonRpcHandler: RpcHandler, THttpApiHandler: Clone + Send + Sync + 'static
     pub fn build(self, starknet_config: &StarknetConfig) -> ServerResult<StarknetDevnetServer> {
         let mut svc = self.routes;
 
+        println!("DEBUG setting request body size: {}", starknet_config.request_body_size_limit);
         svc = svc
             .layer(Extension(self.json_rpc_handler))
             .layer(Extension(self.http_api_handler))
             .layer(TraceLayer::new_for_http())
-            .layer(TimeoutLayer::new(Duration::from_secs(starknet_config.timeout.into())));
+            .layer(TimeoutLayer::new(Duration::from_secs(starknet_config.timeout.into())))
+            .layer(DefaultBodyLimit::max(starknet_config.request_body_size_limit));
 
         if let Some(ServerConfig { allow_origin }) = self.config {
             svc = svc.layer(
--- a/crates/starknet-devnet-server/src/builder.rs
+++ b/crates/starknet-devnet-server/src/builder.rs
@@ -3,12 +3,12 @@ use std::net::SocketAddr;
 use std::time::Duration;
 
 use axum::extract::DefaultBodyLimit;
+use axum::http::HeaderValue;
 use axum::response::Response;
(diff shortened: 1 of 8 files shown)


mikiw (approved):
looks legit

marioiordanov (commented):
with the addition of this new cli param request_body_size_limit, I think its time to split the `timeout` option and this new one into separate config, which are related to configuration on the infrastructure part of the application. The other ones are options for the business logic

crates/starknet-devnet-server/src/builder.rs
```diff
@@ -95,11 +96,13 @@ impl<TJsonRpcHandler: RpcHandler, THttpApiHandler: Clone + Send + Sync + 'static
     pub fn build(self, starknet_config: &StarknetConfig) -> ServerResult<StarknetDevnetServer> {
         let mut svc = self.routes;
 
+        println!("DEBUG setting request body size: {}", starknet_config.request_body_size_limit);
```
marioiordanov, line 99: This print statement will be shown on each startup. Why not log It with `debug!` macro

FabijanC: lol, I accidentally left it and didn't see it in the self-review, thanks for noticing

FabijanC: Removed in https://github.com/0xSpaceShard/starknet-devnet-rs/pull/410/commits/993e799a5f087eb72f3a589c3a2fee2a606deeda

crates/starknet-devnet/src/cli.rs
```diff
@@ -130,17 +131,23 @@ pub(crate) struct Args {
     #[arg(help = "Specify the number of the block to fork at;")]
     #[arg(requires = "fork_network")]
     fork_block: Option<u64>,
+
+    #[arg(long = "request-body-size-limit")]
+    #[arg(value_name = "BYTES")]
+    #[arg(help = "Specify the maximum HTTP request body size;")]
+    #[arg(default_value_t = DEVNET_DEFAULT_REQUEST_BODY_SIZE_LIMIT)]
+    request_body_size_limit: usize,
```
FabijanC, line 139: I'm unsure about using the `usize` type for this, as it is not machine-agnostic.

marioiordanov: No problem, even if its on 32bit machine 4 bytes are 4_294_967_295 which is 4GB +.

FabijanC: True, though I was more interested in having the same behavior on all platforms, but maybe it's not so important

marioiordanov (approved):

PR #430, opened 2024-04-16
Block on demand

mikiw:
## Usage related changes

- add block on demand feature
- dump/load `transaction` mode was renamed to `block` mode

- `pending_state` in `Starknet` was added

## Checklist:

- [x] Checked the relevant parts of [development docs](https://github.com/0xSpaceShard/starknet-devnet-rs/?tab=readme-ov-file#development)
- [x] Updated the docs if needed, including the [TODO section](https://github.com/0xSpaceShard/starknet-devnet-rs/?tab=readme-ov-file#todo-to-reach-feature-parity-with-the-pythonic-devnet)
- [x] Updated the tests if needed; all passing ([execution info](https://github.com/0xSpaceShard/starknet-devnet-rs/?tab=readme-ov-file#test-execution))


mikiw:
as @FabijanC suggested we could change dump mode from transaction to block which makes more sense

crates/starknet-devnet/tests/test_advancing_time.rs
```diff
@@ -259,18 +257,33 @@ mod advancing_time_tests {
...
         tokio::time::sleep(time::Duration::from_secs(1)).await;
 
-        // check if after mint timestamp > last block, check if inside buffer limit
-        devnet.mint(DUMMY_ADDRESS, DUMMY_AMOUNT).await;
-        let mint_block = devnet.get_latest_block_with_tx_hashes().await.unwrap();
-        assert_gt_with_buffer(mint_block.timestamp, empty_block.timestamp);
+        // check if after create block timestamp > last block, check if inside buffer limit
+        devnet.create_block().await.unwrap();
+        let latest_block = devnet.get_latest_block_with_tx_hashes().await.unwrap();
+        assert_gt_with_buffer(latest_block.timestamp, empty_block.timestamp);
```
FabijanC, line 263: So far minting was used just to induce the creation of a new block (with new timestamp)?

crates/starknet-devnet/tests/test_advancing_time.rs
```diff
@@ -259,18 +257,33 @@ mod advancing_time_tests {
...
+
+        set_time_in_past(&devnet).await;
+    }
+
+    #[tokio::test]
+    async fn set_time_in_past_block_on_demand_mode() {
+        let devnet: BackgroundDevnet =
+            BackgroundDevnet::spawn_with_additional_args(&["--blocks-on-demand"])
+                .await
+                .expect("Could not start Devnet");
```
FabijanC, line 278: Here and in other places, the line can be shorter if you just do:
```rust
let devnet = BackgroundDevnet::spawn_with_additional_args(&["--blocks-on-demand"]).await.unwrap();
```

crates/starknet-devnet/tests/test_dump_and_load.rs
```diff
@@ -100,6 +100,49 @@ mod dump_and_load_tests {
...
+                    "--dump-path",
+                    &dump_file.path,
+                    "--dump-on",
+                    mode,
+                    "--blocks-on-demand",
+                ])
+                .await
+                .expect("Could not start Devnet");
+
+                devnet_dump.mint(DUMMY_ADDRESS, DUMMY_AMOUNT).await;
```
FabijanC, line 123: Are there any blocks-on-demand tests that produce transactions that aren't minting? It's crucial for testing IMO

crates/starknet-devnet/tests/test_dump_and_load.rs
```diff
@@ -100,6 +100,49 @@ mod dump_and_load_tests {
...
+                "--dump-on",
+                mode,
+                "--blocks-on-demand",
+            ])
+            .await
+            .expect("Could not start Devnet");
+
+            let last_block = devnet_load.get_latest_block_with_tx_hashes().await.unwrap();
+            assert_eq!(last_block.block_number, 2);
+            assert_eq!(last_block.transactions.len(), 2);
```
FabijanC, line 142: But shouldn't it be 1? 0 is genesis block and 1 is the only new created block (containing two minting txs)?

crates/starknet-devnet/tests/test_get_events.rs
```diff
@@ -43,6 +42,10 @@ mod get_events_integration_tests {
 
         let predeployed_account = Arc::new(predeployed_account);
 
+        if block_on_demand {
+            let _ = devnet.create_block().await;
```
FabijanC, line 46: Here and in other places, I'd avoid `let _ =`, but I'd unwrap:
```rust
devnet.create_block().await.unwrap();
```

crates/starknet-devnet/tests/test_get_events.rs
```diff
@@ -69,15 +76,21 @@ mod get_events_integration_tests {
 
         // invoke 10 times the contract to emit event, it should produce 10 events
         let n_events_contract_invokations = 10;
```
FabijanC, line 78: It should be `invocation`, not `invokation`

crates/starknet-devnet/tests/test_blocks_on_demand.rs
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    #[tokio::test]
+    async fn blocks_on_demand_states_and_blocks() {
+        let devnet: BackgroundDevnet =
+            BackgroundDevnet::spawn_with_additional_args(&["--blocks-on-demand"])
+                .await
+                .expect("Could not start Devnet");
+
+        for _ in 0..TX_COUNT {
```
FabijanC, line 64: I think this tx_count should be a local variable defined inside the test, not that I have to scroll to the top of the file to see it.

FabijanC (changes requested):

README.md
```diff
@@ -238,7 +235,7 @@ POST /mint
 Check the balance of an address by sending a GET request to `/account_balance`. The address should be a 0x-prefixed hex string; the unit defaults to `WEI`.
 
 ```
-GET /account_balance?address=<ADDRESS>&[unit=<FRI|WEI>]
+GET /account_balance?address=<ADDRESS>&[unit=<FRI|WEI>]&[pending_state=<true|false>]
```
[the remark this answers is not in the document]
mikiw: will be changed to BlockTag

mikiw:
Add logic and test for querying pending block with normal mode

README.md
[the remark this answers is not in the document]
mikiw: so I tried to implement impl std::fmt::Display for TestTag, where TestTag was a pub struct TestTag(BlockTag) but it lead to "cannot initialize a tuple struct which contains private field" problem

```
use starknet_rs_core::types::BlockTag;

pub struct TestTag(BlockTag);

impl std::fmt::Display for TestTag {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        f.write_str(match &self {
            TestTag(BlockTag::Latest) => "latest",
            TestTag(BlockTag::Pending) => "pending",
        })
    }
}
```

this is why it led to this match code inside function
```
        let tag = match &self {
            BlockTag::Latest => "latest",
            BlockTag::Pending => "pending",
        };
```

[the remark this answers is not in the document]
mikiw: Fixed in da7756e07371af939546838a7b014c120ce52644 but we can discuss `impl std::fmt::Display` further

README.md
```diff
@@ -280,14 +277,14 @@ $ starknet-devnet --dump-path <PATH>
...
+Currently, dumping produces a list of received transactions that are stored on disk.
 Conversely, loading is implemented as the re-execution of transactions from a dump.
 This means that timestamps of `StarknetBlock` will be different.
 
 ### Loading disclaimer
 
-Dumping and loading is not guaranteed to work cross-version. I.e. if you dumped one version of Devnet, do not expect it to be loadable with a different version.
-If you dumped a Devnet utilizing one class for account predeployment (e.g. the default `--account-class cairo0`), you should use the same option when loading.
+Dumping and loading are not guaranteed to work cross-version. I.e. if you dumped one version of Devnet, do not expect it to be loadable with a different version.
+If you dumped a Devnet utilizing one class for account predeployment or with blocks-on-demand mode (e.g. the default `--account-class cairo0` and/or `--blocks-on-demand`), you should use the same options when loading.
```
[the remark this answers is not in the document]
mikiw: Fixed in 52b1f86b4ce8bc791299a7cb29b6d11c913f4262

README.md
```diff
@@ -301,6 +298,26 @@ Devnet starts with a genesis block (with a block number equal to 0). In forking
 
 A new block is generated with each new transaction, and you can create an empty block by yourself.
 
+### Creating blocks on demand
+
+If you start Devnet with the `--blocks-on-demand` CLI option, all valid transactions will be stored in a pending block (targetable via block tag `"Pending"`).
+
+To create a block on demand, send a `POST` request to `/create_block`. This will convert the pending block to the latest block (targetable via block tag `"Latest"`), giving it a block hash and a block number. All subsequent transactions will be stored in a new pending block.
```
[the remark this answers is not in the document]
mikiw: Fixed in 52b1f86b4ce8bc791299a7cb29b6d11c913f4262

crates/starknet-devnet-core/src/blocks/mod.rs
```diff
@@ -368,11 +370,17 @@ mod tests {
...
             block_to_insert.header.block_number = BlockNumber(block_number);
             block_to_insert.header.block_hash = Felt::from(block_number as u128).into();
-            blocks.insert(block_to_insert, StateDiff::default());
+            blocks.insert(block_to_insert.clone(), StateDiff::default());
+
+            // last block will be a pending block
+            if block_number == last_block_number - 1 {
+                blocks.pending_block = block_to_insert;
+            }
         }
```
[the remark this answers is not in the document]
mikiw: Fixed in 5397253f8b1160a688b3e70a6ac70d5f8d2eceb3

crates/starknet-devnet-core/src/starknet/add_invoke_transaction.rs
```diff
@@ -38,11 +38,17 @@ pub fn add_invoke_transaction(
         }
     };
 
+    let state = if starknet.config.blocks_on_demand {
+        &mut starknet.pending_state.state
+    } else {
+        &mut starknet.state.state
+    };
```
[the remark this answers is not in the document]
mikiw: ok, will fix, I didn't know that block_context can be cloned.

crates/starknet-devnet-core/src/blocks/mod.rs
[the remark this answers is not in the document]
FabijanC: No, I specifically meant `2..=last_block_number`, not `2..last_block_number + 1` (notice the `=` in my version). I think it's more straightforward that way (if one is acquainted with the `..=` syntax)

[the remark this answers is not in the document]
mikiw: but with that test will not pass

crates/starknet-devnet/tests/test_get_events.rs
mikiw: 32178c97ddfa6f276c82bede3bf8d7d6ffba1e73

crates/starknet-devnet/tests/test_blocks_on_demand.rs
mikiw: 43cfd9b0e2946fc54f7d4c073bc26c201087fb0f

crates/starknet-devnet/tests/test_get_events.rs
mikiw: bf091b7494899019ace95a45775796b445133716

crates/starknet-devnet-core/src/blocks/mod.rs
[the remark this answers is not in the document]
mikiw: I didn't notices this `=`, fixed in 3407094437a098a658814b6299b2eca8ab206d96

crates/starknet-devnet/tests/test_advancing_time.rs
mikiw: remove of redundant : BackgroundDevnet in whole project is here 5916ecc01a0cf4ef00799b29814e9daa662bf93a

crates/starknet-devnet/tests/test_advancing_time.rs
mikiw: Yes this is why I changed it to create_block()

crates/starknet-devnet/tests/test_dump_and_load.rs
mikiw: I can introduce test like this

crates/starknet-devnet/tests/test_dump_and_load.rs
mikiw: There is loop with `for _ in 0..2`.

This loop was needed to test all scenarios. As far as I remember it was related to doublets.

crates/starknet-devnet/tests/common/background_devnet.rs
```diff
@@ -223,12 +223,25 @@ impl BackgroundDevnet {
...
         address: &FieldElement,
         unit: FeeUnit,
     ) -> Result<FieldElement, anyhow::Error> {
-        let params = format!("address={:#x}&unit={}", address, unit);
+        Self::get_balance_pending_state(self, address, unit, false).await
+    }
+
+    /// Get balance at contract_address, as written in the ERC20 contract corresponding to `unit`
+    /// from pending state or latest state
+    pub async fn get_balance_pending_state(
```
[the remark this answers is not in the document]
mikiw: after refactoring it was renamed to get_balance_by_tag

887ba2bc65f77a8ef2826466587ac15984ebc538

crates/starknet-devnet/tests/test_dump_and_load.rs
FabijanC: I completely forgot it's inside the loop, it's correct then. Just, in that case, I'd suggest extracting the `2` to a variable (`total_iterations` or something like that) which is used both in loop definition and in assertions.

README.md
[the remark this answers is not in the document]
mikiw: Fixed in 2185687511e9a54007f67bca3002b4ed2943915c

crates/starknet-devnet/tests/test_dump_and_load.rs
mikiw: fixed in 9a96b6e9126187cc44c2d4e2f70163dfb494a8b3

mikiw:
> Add logic and test for querying pending block with normal mode

added in 35be702cba97a1e8f41fbf306f271de875722239

crates/starknet-devnet/tests/test_dump_and_load.rs
mikiw: Added in 414a3adb369ff7b50388c43fa457ec5fed3d4814

crates/starknet-devnet/tests/test_advancing_time.rs
FabijanC: > "Could not start Devnet" appears in my terminal so I know that I need to build a release

If `.expect(...)` is used, you get:
```
Could not start devnet: DevnetNotStartable
```

If `.unwrap()` is used, you get:
```
called `Result::unwrap()` on an `Err` value: DevnetNotStartable
```

Just for future spawnings of BackgroundDevnet. I'll resolve this conversation.

crates/starknet-devnet-server/src/api/json_rpc/endpoints.rs
```diff
@@ -23,14 +23,20 @@ impl JsonRpcHandler {
...
-            )?;
+    pub async fn get_block_with_tx_hashes(&self, mut block_id: BlockId) -> StrictRpcResult {
+        let starknet = self.api.starknet.read().await;
+
+        if !starknet.config.blocks_on_demand
+            && block_id == BlockId(ImportedBlockId::Tag(BlockTag::Pending))
+        {
+            // in normal mode, querying pending block should default to the latest
+            block_id = BlockId(ImportedBlockId::Tag(BlockTag::Latest));
+        }
```
FabijanC, line 34: It is easier to have a conversation in comments than in posts on PR. Continuing from https://github.com/0xSpaceShard/starknet-devnet-rs/pull/430#issuecomment-2076689281

I feel like 35be702cba97a1e8f41fbf306f271de875722239 only partly addresses the original issue. Shouldn't we always default to latest if no pending, for all rpc calls? Also, I feel like this block selection should be done as deep as possible, e.g. in `get_by_block_id`

mikiw: In `Starknet` struct there is a `restart_pending_block()` which is resetting `self.blocks.pending_block` so pending_block is always created and updated so there is no way of detecting if there is one or not.

And about `get_by_block_id()` - in this layer we don't have information about `starknet.config.blocks_on_demand` because it's inside`StarknetBlocks`.

crates/starknet-devnet/tests/common/background_devnet.rs
```diff
@@ -225,12 +225,24 @@ impl BackgroundDevnet {
...
+
+    /// Get balance at contract_address, as written in the ERC20 contract corresponding to `unit`
+    /// from pending state or latest state
+    pub async fn get_balance_by_tag(
+        &self,
+        address: &FieldElement,
+        unit: FeeUnit,
+        tag: BlockTag,
+    ) -> Result<FieldElement, anyhow::Error> {
+        let params = format!("address={:#x}&unit={}&tag={}", address, unit, Self::tag_to_str(tag));
```
FabijanC, line 245: Something doesn't feel right here. We now have: `get_balance`, `get_balance_by_block` and `get_balance_by_tag`. The functionality of `get_balance_by_tag` is a subset of the functionality of `get_balance_by_block`. We can rename `get_balance` to `get_latest_balance` (or `get_balance_latest`).

The confusion grows even more with the introduction of `get_contract_balance` in L262. IMO that method doesn't belong to `BackgroundDevnet`, it's a utility method that should be better documented (explaining what exactly is this balance - not the usual ERC20 balance), and used as a util function (defined outside of BackgroundDevnet).

crates/starknet-devnet-server/src/api/json_rpc/endpoints.rs
FabijanC: Would our problems be solved if pending block became `Option`?

mikiw: probably yes but this would require getting rid of `restart_pending_block()` logic and it's not the scope of this ticket

FabijanC: Making block_id=`pending` default to `latest` is the scope of this ticket, so we're in a dilemma

mikiw: The current code is defaulting to the `latest`, the problem is that the current business logic is using `restart_pending_block()` and this should be removed.

What I may suggest is to open a new ticket with the refactoring of restart_pending_block() and making `self.blocks.pending_block` optional.

mikiw: https://github.com/0xSpaceShard/starknet-devnet-rs/issues/445

crates/starknet-devnet/tests/common/background_devnet.rs
mikiw: 3ebad46f679dfbad34ca9e862352e506a5e2ea27

crates/starknet-devnet/tests/test_blocks_on_demand.rs
```diff
@@ -0,0 +1,175 @@
...
+            .await
+            .unwrap();
+
+        devnet.create_block().await.unwrap();
+
+        assert_tx_successful(&invoke_result.transaction_hash, &devnet.json_rpc_client).await;
+        assert_eq!(
+            get_contract_balance(&devnet, contract_address).await,
+            initial_value + increment
+        );
```
FabijanC, line 173: Can we also assert here that all of this resulted in one block? and that the block contains only the tx hashes corresponding to declare, deploy, invoke

mikiw: I'm not sure if you can declare, deploy, and invoke with in one block in Starknet

FabijanC: I think this is a very important part of the pending state/block functionality, this needs to be discussed in more detail

FabijanC (changes requested):

crates/starknet-devnet-server/src/api/json_rpc/endpoints.rs
mikiw: I'll check that

mikiw: fixed in 16004864730e942e84f8a0371b8f44ec989a717a, is not that bad with blocks_on_demand inside StarknetBlocks, actually it makes sense

crates/starknet-devnet/tests/test_blocks_on_demand.rs
mikiw: discussed on a call I added additional invoke here 7f5f5c7db5a702c5e0e3e776b1eddbacf6d6c500

crates/starknet-devnet-core/src/blocks/mod.rs
```diff
@@ -86,12 +88,17 @@ impl StarknetBlocks {
...
+            // in normal mode, querying pending block should default to the latest
+            block_id = &BlockId::Tag(BlockTag::Latest);
+        }
+
         match block_id {
             BlockId::Hash(hash) => self.get_by_hash(Felt::from(hash)),
             BlockId::Number(block_number) => self.get_by_num(&BlockNumber(*block_number)),
-            // latest and pending for now will return the latest one
-            BlockId::Tag(_) => {
+            BlockId::Tag(BlockTag::Pending) => Some(&self.pending_block),
```
FabijanC, line 100: Instead of requiring a mutable `block_id`, I think it's clearer if the `!self.blocks_on_demand` check is done inside the `BlockId::Tag(BlockTag::Pending)` arm of the `match`

crates/starknet-devnet/tests/test_blocks_on_demand.rs
```diff
@@ -0,0 +1,185 @@
...
+            .await;
+
+        assert_latest_block_with_transactions(&devnet, 1, tx_hashes).await;
+
+        // check if pending_block was restarted
+        let pending_block = devnet.get_pending_block_with_tx_hashes().await.unwrap();
+        assert_eq!(pending_block.block_number, 2);
+        assert_eq!(pending_block.transactions.len(), 0);
+
+        assert_pending_block_with_transactions(&devnet, 2, 0).await;
```
FabijanC, line 90: Why are these two assertions done both here and in `assert_pending_block_with_transactions`? If this is the only place where the function is used, it's ok to remove it. Also, perhaps better to assert tx hash than tx count

mikiw: Right, good that you spotted that, somehow I missed that. My mistake, fixed in b8fc6af3cac957f64a722c0a872c57e473a4fa18

crates/starknet-devnet-core/src/blocks/mod.rs
mikiw: e8a7f3b87565b0e7e949e89980fe50186bc744c9

PR #437, opened 2024-04-23
chore: removed unnecessary derive attributes in the codebase

fishonamos:
Determined the unnecessary / necessary derive attributes in the codebase and removed them. This will solve #188


FabijanC (changes requested):
The project is not compilable. The PR has been open for more than a week, and it has served no practical purpose, so the safest thing to do is to close it. Prove me wrong and I will reopen it.

PR #443, opened 2024-04-25
feat: add codespell workflow for CI

### LINKED ISSUE #258: feat: add codespell linter workflow for CI
Codespell linter tool is very important for code and system quality, especially for enhancing docs readability.

So I want to add codespell linter workflow for CI.

xiaolou86:
Closes #258 
## Usage related changes

None

Add codespell check workflow for CI
## Checklist:

- [ ] Checked the relevant parts of [development docs](https://github.com/0xSpaceShard/starknet-devnet-rs/?tab=readme-ov-file#development)
- [ ] Updated the docs if needed, including the [TODO section](https://github.com/0xSpaceShard/starknet-devnet-rs/?tab=readme-ov-file#todo-to-reach-feature-parity-with-the-pythonic-devnet)
- [ ] Linked the [issues](https://github.com/0xSpaceShard/starknet-devnet-rs/issues) which this PR resolves
- [ ] Updated the tests if needed; all passing ([execution info](https://github.com/0xSpaceShard/starknet-devnet-rs/?tab=readme-ov-file#test-execution))

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -69,6 +69,9 @@ jobs:
       - run:
           name: Format check
           command: cargo +nightly fmt --all --check
+      - run:
+          name: Code spell check
+          command: cargo +nightly install typos-cli; typos
       - save_cache:
           # Caching dependencies for future use
           # Build and Clippy steps generate their own target artifacts
@@ -90,18 +93,6 @@ jobs:
           name: Integration tests
           # fails if jobs not limited; read more in https://github.com/0xSpaceShard/starknet-devnet-rs/issues/378
           command: cargo test --jobs 3 --test '*' --no-fail-fast
-
-  spell-check:
-    docker:
-      - image: cimg/rust:1.75.0
-    steps:
-      - checkout
-      - run:
-          name: Install spelling checker
-          command: cargo install typos-cli
-      - run:
-          name: Execute spelling checker
-          command: typos
 
   binary-build:
     parameters:
@@ -186,7 +177,6 @@ workflows:
     jobs:
       - build-and-test:
           cargo_cache_key: cargo-cache-{{ checksum "Cargo.lock" }}-{{ checksum "rust-toolchain.toml" }}-v1
-      - spell-check
       - binary-build:
           <<: *on_main
           matrix:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -71,7 +71,7 @@ jobs:
           command: cargo +nightly fmt --all --check
       - run:
           name: Code spell check
-          command: cargo +nightly install typos-cli; typos
+          command: cargo +nightly install typos-cli && typos
       - save_cache:
           # Caching dependencies for future use
           # Build and Clippy steps generate their own target artifacts
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -41,17 +41,6 @@ executors:
 # Define a job to be invoked later in a workflow.
 # See: https://circleci.com/docs/configuration-reference/#jobs
 jobs:
-  spell-check:
-    docker:
-      - image: cimg/rust:1.75.0
-    steps:
-      # Checkout the code as the first step.
-      - checkout
-      # Install typos-cli
-      - run: cargo install typos-cli
-      # Run code spell check
-      - run: typos
-
   build-and-test:
     parameters:
       cargo_cache_key:
@@ -101,6 +90,18 @@ jobs:
           name: Integration tests
           # fails if jobs not limited; read more in https://github.com/0xSpaceShard/starknet-devnet-rs/issues/378
           command: cargo test --jobs 3 --test '*' --no-fail-fast
+
+  spell-check:
+    docker:
+      - image: cimg/rust:1.75.0
+    steps:
+      - checkout
+      - run:
+          name: Install spelling checker
+          command: cargo install typos-cli
+      - run:
+          name: Execute spelling checker
+          command: typos
 
   binary-build:
     parameters:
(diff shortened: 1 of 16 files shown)


FabijanC:
I added a commit hoping that it would trigger CI, it didn't, I'll try something else, stay tuned

.circleci/config.yml
```diff
@@ -91,6 +91,18 @@ jobs:
           # fails if jobs not limited; read more in https://github.com/0xSpaceShard/starknet-devnet-rs/issues/378
           command: cargo test --jobs 3 --test '*' --no-fail-fast
 
+  spell-check:
+    docker:
+      - image: cimg/rust:1.75.0
```
FabijanC, line 96: I'm skeptical about this job being the only one dependent on rust 1.75.0 and everything else using 1.74.0. I see that typos-cli 1.16.26 is the last version using a rust version before 1.75.0 (more specifically, it's dependent on rust 1.70.0), so one option is installing that version, just not sure how much behind with all the features that version is.

The other option is to update the whole toolchain to 1.75.0, but not sure if we want to do that just to support a spelling checker.

The list of versions with rust requirements: https://crates.io/crates/typos-cli/versions

xiaolou86:
Is it because the PR comes from external forked repo?

.circleci/config.yml
FabijanC: Locally I ran this:
```
docker run --rm -it --mount type=bind,source=/home/fabijanc/path/to/starknet-devnet-rs/,target=/home/circleci/project cimg/rust:1.75.0
```

and inside it I ran:
```
$ cargo install typos-cli
$ typos
```

and got this:
```
error: failed to compile `typos-cli v1.20.10`, intermediate artifacts can be found at `/tmp/cargo-install94t5c8`.
To reuse those artifacts with a future compilation, set the environment variable `CARGO_TARGET_DIR` to that path.

Caused by:
  package `typos-vars v0.8.16` cannot be built because it requires rustc 1.75 or newer, while the currently active rustc version is 1.74.0
  Try re-running cargo install with `--locked`
```

(Installing with `--locked` doesn't help)

FabijanC (changes requested):

_typos.toml
```diff
@@ -0,0 +1,14 @@
...
+addresss = "addresss"
+
+# test for incorrect block_number key in deserialize_block_id_block_number_variants
+numberr = "numberr"
+
+# STARKNET_1.0_MSGING_L1TOL2_MAPPPING_V2
+MAPPPING = "MAPPPING"
+
+payed = "payed"
+ede = "ede"
```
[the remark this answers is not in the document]
FabijanC: Ok I may have found the answer in the [docs](https://crates.io/crates/typos) under `False positives`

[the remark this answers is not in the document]
FabijanC: Perhaps the entire submodule can be ignored, to prevent false negatives (i.e. somebody writing `payed` outside of the submodule and it not being registered as an error)

FabijanC:
I was able to trigger the workflow, it failed. Be sure to go through the checkboxes in your PR description and through my comments.

_typos.toml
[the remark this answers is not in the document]
xiaolou86: Ok, it's more reasonable.
1. I will ignore the entire directory "contracts/l1-l2-messaging/solidity/" instead just ignore `MAPPPING` and `payed`. I think it's a submodule?
2. `addressss` and `numberr` are constructed as incorrect keys on purpose for test cases in `crates/starknet-devnet-server/src/api/json_rpc/models.rs`. Maybe it's better to just ignore this file.

xiaolou86:
Got it, thank you for your quickly review!

.circleci/config.yml
xiaolou86: I find that it is because the `rust-toolchain.toml` file will override the version.

FabijanC: makes sense, to reduce the impact of your PR, can you try installing the mentioned older version of typos-cli (1.16.26)?

xiaolou86:
I updated with the following changes:
1. Use existing nightly toolchain instead of installing v1.75.0.
2. Don't correct submodule typo.
3. Use mock key to avoid typo error.

@FabijanC Please help me to review again. Thank you very much.

.circleci/config.yml
FabijanC: Resolved by using the nightly toolchain

_typos.toml
```diff
@@ -0,0 +1,7 @@
+[files]
+# Don't correct submodule and json files
+extend-exclude = ["contracts/l1-l2-messaging/solidity/*", "*.json"]
+
+[default.extend-words]
+# Don't correct the word
+ede = "ede"
```
FabijanC, line 7: Is this character sequence only encountered in hexadecimal strings? Is there really no smarter way around it than ignoring it?

One example of a false negative: someone wants to write "edge", but writes "ede" - this is not registered, right?

FabijanC: I think I found a solution. What do you think about the following:
```toml
[files]
# Don't correct submodule and json files
extend-exclude = ["contracts/l1-l2-messaging/solidity", "crates/starknet-devnet-server/test_data"]
ignore-vcs = true

[default]
extend-ignore-re = ["0x[0-9a-fA-F]+"]
```

I added the regex, and `ignore-vcs`, and also modified the `extend-exclude` paths a bit.

FabijanC: more options: https://github.com/crate-ci/typos/blob/master/docs/reference.md

xiaolou86: Awesome! I will try.

FabijanC (approved):

Issue #342, opened 2024-02-12
Remove Copy trait from Felt


marioiordanov:
Copy trait must be removed from 
```rust
#[derive(Debug, Default, Copy, Clone, Eq, PartialEq, Hash, PartialOrd, Ord)]
pub struct Felt(pub(crate) [u8; 32]);
```

marioiordanov:
To avoid copying 32 bytes of data when the idea is just to move the element

g4titanx:
this looks interesting!
i would love to work on this

ivpavici:
@g4titanx please report frequently on progress!

g4titanx:
okay

g4titanx:
> Copy trait must be removed from
> 
> ```rust
> #[derive(Debug, Default, Copy, Clone, Eq, PartialEq, Hash, PartialOrd, Ord)]
> pub struct Felt(pub(crate) [u8; 32]);
> ```

where is the fely struct declared

g4titanx:
Of course, using a Vec<&Felt> for calldata_to_hash can be a solution, but it caused a series errors in the program, just like adding the cloned() method on line 693. I'm currently stuck on this issue, and I feel like I'm missing something important.
i would be so glad if you can offer an helping hand

g4titanx:
Hello @ivpavici, any update on this

marioiordanov:
@g4titanx in fact this issue is not an easy one the idea is to remove the copying of 32 bytes, where moving of the variable was intended.

g4titanx:
i agree. it's not as easy as I thought. but, I will give it all I can

g4titanx:
> hello @ivpavici, as expected removing the copy trait from Felt led to a chain of issues. some of the errors I got during compilation were from the custom types that implemented the copy trait with the Felt type in their field. i had to remove their copy traits tho.
> 
> now one error I am stuck with is "type mismatch resolving `<Iter<'_, Felt> as IntoIterator>::Item == Felt` " in this section of transaction.rs
> 
> https://github.com/0xSpaceShard/starknet-devnet-rs/blob/9162df948ffc444520ba1c7397d9e8564e8dc576/crates/starknet-devnet-types/src/rpc/transactions.rs#L692-L693

tried different approach to fixing this issue.
i am open to ideas how I can advance.

ivpavici:
@g4titanx can you open a PR with your approach? It's hard to tell like this what you tried

[...]

PR #457, opened 2024-05-01
removed copy trait from Felt 

g4titanx:
i opened the PR as suggested. you can see the approach i took in fixing this issue: https://github.com/0xSpaceShard/starknet-devnet-rs/issues/342

### DIFF
--- a/crates/starknet-devnet-types/src/contract_storage_key.rs
+++ b/crates/starknet-devnet-types/src/contract_storage_key.rs
@@ -3,7 +3,7 @@ use std::fmt::Display;
 use super::contract_address::ContractAddress;
 use crate::patricia_key::StorageKey;
 
-#[derive(Debug, Default, Clone, Copy, PartialEq, Eq, Hash)]
+#[derive(Debug, Default, Clone, PartialEq, Eq, Hash)]
 pub struct ContractStorageKey(ContractAddress, StorageKey);
 
 impl ContractStorageKey {
--- a/crates/starknet-devnet-types/src/patricia_key.rs
+++ b/crates/starknet-devnet-types/src/patricia_key.rs
@@ -7,7 +7,7 @@ use crate::serde_helpers::hex_string::{
     deserialize_to_prefixed_patricia_key, serialize_patricia_key_to_prefixed_hex,
 };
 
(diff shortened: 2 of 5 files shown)


g4titanx:
here i am currently trying to fix this error:
error[E0271]: type mismatch resolving `<Iter<'_, Felt> as IntoIterator>::Item == Felt`
   --> crates/starknet-devnet-types/src/rpc/transactions.rs:693:41
    |
693 |                 calldata_to_hash.extend(v1.constructor_calldata.iter());
    |                                  ------ ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `Felt`, found `&Felt`
    |                                  |
    |                                  required by a bound introduced by this call
    |

my first approach in fixing this error was to do this: 
                let mut calldata_to_hash = vec![&v1.class_hash, &v1.contract_address_salt];
now, this fixes this particular error but at the sam time, it generates more errors across different crates. here's an example:
error[E0507]: cannot move out of `v3.common.version` which is behind a shared reference
   --> crates/starknet-devnet-types/src/rpc/transactions.rs:880:21
    |
880 |                     v3.common.version,
    |                     ^^^^^^^^^^^^^^^^^ move occurs because `v3.common.version` has type `Felt`, which does not implement the `Copy` trait
    
now, what do you suggest

g4titanx:
i have tried a copy of metthods, that was about a week ago. 
honestly, I cant remember everything I tried. but, I know the one I mentioned was one of the best approach I could come up with. 

also, another way I tried was to create a reference through each crate since Felt doesn't implement a copy trait. which led me to creating a lifetime for each of the reference, it was too complemented. oh, it fixed most of the errors but at the end, it was going to increase the compile time and it led to one or two errors i couldn't fix.

how would you fix this?

FabijanC (changes requested):
When you remove `Copy`, it means you'll have to add more `.clone()`

g4titanx:
> When you remove `Copy`, it means you'll have to add more `.clone()`

okay, I would do that

marioiordanov:
IMO solving this issue will require bigger refactoring, either use references or move the variables where needed. Mainly depends on the context of the caller

PR #499, opened 2024-06-18
Introduce global RPC class storage

### LINKED ISSUE #272: Memory exhaustion
**Describe the bug (observed vs expected behavior)**

We completed our migration from devnet python to devnet rust, after a few adjustments now we have all tests running again. There is only one issue. Our tests take around 20 min to run, during that time the devnet memory keeps increasing, after 5 minutes it reaches 7.99Gb and it starts to slow down, after another minute the container stops (no specific error i could find)

As a workaround we are calling the `restart` endpoint every 2 minutes, and that seems to help, although i still see the memory increasing it’s good enough for our test suite (edited)

**Not reproducible on alpha-goerli**

- [x] This issue is only present on Devnet and cannot be reproduced on alpha-goerli (check the box if true).

**To Reproduce**

1. Keep using devnet rust for 20 minutes

**Devnet version**

- I am using Devnet version: shardlabs/starknet-devnet-rs:64c425b832b96ba09b49646fe0fbf49862c0fb6d
- [x] This happens with a dockerized Devnet (check the box if true).
- This does not appear on the following Devnet version:

**System specifications**

- OS: MacOS and also running Github actions on ubuntu

FabijanC:
## Usage related changes

- Close #272 
- Implement the proposal described in https://github.com/0xSpaceShard/starknet-devnet-rs/issues/272#issuecomment-1924033491

## Benchmarking
- Running `./scripts/benchmark/command_stat_test.py` (500 mints):
- comparing Devnet on `main` and Devnet on this PR's branch
- Devnet binaries compiled with `cargo build --release`
- sample size: 10
- alternative hypothesis: two-sided

### With `--state-archive-capacity full`
```
Time (s):   TtestResult(statistic=-2.102220212750287, pvalue=0.04987296605083779, df=18.0)
  - Mean (main vs this branch): 6.6 s vs 7.0 s
Memory (MB): TtestResult(statistic=853.6179541610356, pvalue=6.351731250129005e-43, df=18.0)
  - Mean (main vs this branch): 997 MB vs 34 MB
```

### With no feature flags
```
Time (s):   TtestResult(statistic=-1.6918289270796756, pvalue=0.10791760227433438, df=18.0)
  - Mean (main vs this branch): 6.4 s vs 6.7 s
Memory (MB): TtestResult(statistic=7.890611697084889, pvalue=2.979935622929167e-07, df=18.0)
  - Mean (main vs this branch): 20 MB vs 14 MB
```

### Conclusion
- Time slightly worse in regular mode
- Memory consumption reduced in regular and full-state-archive-capacity mode, significantly in the latter.

- Add property to `StarknetState`
  - `rpc_contract_classes: Arc<RwLock<CommittedClassStorage>>`
  - For every class, we record when it was added.
- Main reading logic is in `crates/starknet-devnet-core/src/starknet/get_class_impls.rs`
- Add dependency: `parking_lot`
  - Using its RwLock because it doesn't have to be handled like `std`'s or `tokio`'s
- Expand testing of state update and trace
  - `blocks_on_demand_declarations` fails on `main`
- StateUpdate of blocks in block-on-demand mode was wrong:
  - Bug: always applying empty `StateDiff` in `create_block`.
  - Fixed by accumulating state diffs in a new property called `pending_state_diff` by relying on the newly added `StateDiff::extend`.
- `commit_with_diff` renamed to `commit_diff`, no longer accepts the `state_diff` parameter.
- Extract global contract class creation to `default_global_contract_cache`.
- `StateDiff::generate` accepts newly declared classes instead of the whole class storage.
  - This means that the responsibility of extracting these classes has been moved to the caller.
- `assert_cairo1_classes_equal` now working with references - removes unnecessary cloning.

## Checklist:

- [x] Updated the tests if needed; all passing - [execution info](https://github.com/0xSpaceShard/starknet-devnet-rs/?tab=readme-ov-file#test-execution)


crates/starknet-devnet-core/src/state/mod.rs
```diff
@@ -285,8 +297,38 @@ impl CustomStateReader for StarknetState {
...
     }
 
-    fn get_rpc_contract_class(&self, class_hash: &ClassHash) -> Option<&ContractClass> {
-        self.rpc_contract_classes.committed.get(class_hash)
+    fn get_rpc_contract_class(
+        &self,
+        class_hash: &ClassHash,
+        block_id: &BlockId,
+    ) -> Option<ContractClass> {
+        let class_storage = self.rpc_contract_classes.read();
```
marioiordanov, line 305: The if -> else if statement makes it hard to understand what is the behavior for BlockId::Pending. Also this function is public and can be called from everywhere, imo better approach is to have an enum (for example NumberOrTag) that represents the possible variants of block_id (number, tag) argument and be able to convert from BlockId to "NumberOrTag". This way you will enforce at compile time that BlockId::Hash is not provided

marioiordanov (changes requested):

crates/starknet-devnet-core/src/state/mod.rs
FabijanC: Addressed in https://github.com/0xSpaceShard/starknet-devnet-rs/pull/499/commits/40b825468d3d2ba9c6af390c765dc916e2be4215

crates/starknet-devnet-core/src/state/mod.rs
```diff
@@ -97,30 +116,31 @@ impl Default for StarknetState {
...
         let new_historic = self.expand_historic(diff.clone())?;
-        self.state = CachedState::new(
-            new_historic.clone(),
-            GlobalContractCache::new(GLOBAL_CONTRACT_CACHE_SIZE_FOR_TEST),
-        );
+        self.state = CachedState::new(new_historic.clone(), default_global_contract_cache());
         Ok(diff)
     }
 
+    pub(crate) fn commit_classes(&mut self, block_number: u64) {
```
mikiw, line 140: Here and in other places why are we using block numbers instead of block hashes? Won't this threaten the abortion block functionality? Tests with abortion block functionality are missing.

crates/starknet-devnet/tests/test_get_class.rs
```diff
@@ -133,4 +133,148 @@ mod get_class_tests {
...
+                devnet.json_rpc_client.get_class(block_id, declaration_result.class_hash).await;
+            match retrieved {
+                Err(ProviderError::StarknetError(StarknetError::ClassHashNotFound)) => (),
+                other => panic!("Unexpected response: {other:?}"),
+            }
+        }
+    }
+
+    #[tokio::test]
+    async fn test_getting_class_with_blocks_on_demand() {
```
mikiw, line 197: shouldn't this test be moved to crates/starknet-devnet/tests/test_blocks_generation.rs
?

crates/starknet-devnet-core/src/state/mod.rs
```diff
@@ -55,39 +67,46 @@ pub trait CustomState {
...
         diff
     }
 }
 
 pub struct StarknetState {
     pub(crate) state: CachedState<DictState>,
-    rpc_contract_classes: CommittedClassStorage,
+    /// The class storage is meant to be shared between states to prevent copying (due to memory
+    /// concerns). Knowing which class was added when is made possible by storing the class
+    /// together with the block number.
```
mikiw, line 95: Why do we use block numbers instead of block hashes?

mikiw (changes requested):

crates/starknet-devnet-core/src/starknet/get_class_impls.rs
```diff
@@ -29,8 +29,21 @@ pub fn get_class_impl(
     block_id: &BlockId,
     class_hash: ClassHash,
 ) -> DevnetResult<ContractClass> {
-    let state = starknet.get_mut_state_at(block_id)?;
-    match state.get_rpc_contract_class(&class_hash) {
+    // if provided with block hash, convert to number - the underlying logic only works with that or
+    // block tag
+    let block_id = if let BlockId::Hash(block_hash) = block_id {
```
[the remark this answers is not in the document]
FabijanC: So that we can query a class storage with `N` classes in `O(1)` by relying on the `if storage_block_number <= query_block_number` trick. Otherwise, for each query we would need to traverse all class commitments, leading to `O(N)`.

crates/starknet-devnet-core/src/starknet/get_class_impls.rs
```diff
@@ -29,8 +29,21 @@ pub fn get_class_impl(
...
+        match starknet.blocks.hash_to_block.get(&block_hash.into()) {
+            Some(block) => BlockId::Number(block.block_number().0),
+            None => return Err(Error::NoBlock),
+        }
+    } else {
+        *block_id
+    };
+
+    // TODO do we even need the state at the specific block_id? Perhaps the class storage should be
```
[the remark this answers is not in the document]
FabijanC: the answer to the question is "probably no", but it would require another refactor (I don't know how big)

crates/starknet-devnet/tests/test_get_class.rs
FabijanC: it makes sense to have it in either of those files

crates/starknet-devnet-core/src/state/mod.rs
FabijanC: I'll have to think more about the impact on block abortion

crates/starknet-devnet-core/src/state/mod.rs
FabijanC: Answered in https://github.com/0xSpaceShard/starknet-devnet-rs/pull/499#discussion_r1650838115

crates/starknet-devnet-core/src/starknet/get_class_impls.rs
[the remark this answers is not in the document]
mikiw: ok got it but still not sure how it will impact block abortion feature

crates/starknet-devnet-core/src/starknet/get_class_impls.rs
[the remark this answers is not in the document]
FabijanC: Refactored and removed TODO in https://github.com/0xSpaceShard/starknet-devnet-rs/pull/499/commits/6a0c35f012a4ba4424dcf6e64a8f0e8e88a2b00a

crates/starknet-devnet-core/src/state/mod.rs
FabijanC: Properly handling block abortion in https://github.com/0xSpaceShard/starknet-devnet-rs/pull/499/commits/ba6f944073fa50dee1c369396c854ffaeb168b1f

crates/starknet-devnet-core/src/starknet/mod.rs
```diff
@@ -815,24 +826,18 @@ impl Starknet {
...
 
-        let genesis_block_number = if let Some(block_number) = self.config.fork_config.block_number
-        {
-            block_number + 1
-        } else {
-            DEVNET_DEFAULT_STARTING_BLOCK_NUMBER
-        };
-        let genesis_block =
-            self.blocks.get_by_block_id(&BlockId::Number(genesis_block_number)).unwrap();
+        let genesis_block = self
```
mikiw, line 837: Is this change related to class storage or we had a bug in block numbers? Is this change here because of forking?

crates/starknet-devnet-core/src/blocks/mod.rs
```diff
@@ -27,6 +27,7 @@ pub(crate) struct StarknetBlocks {
     pub(crate) hash_to_state_diff: HashMap<BlockHash, StateDiff>,
     pub(crate) hash_to_state: HashMap<BlockHash, StarknetState>,
     pub(crate) aborted_blocks: Vec<Felt>,
+    pub(crate) starting_block_number: u64,
```
mikiw, line 30: why do we need to introduce `starting_block_number` seems that it's just `DEVNET_DEFAULT_STARTING_BLOCK_NUMBER` and it never gets updated

crates/starknet-devnet-core/src/state/mod.rs
```diff
@@ -55,72 +61,146 @@ pub trait CustomState {
...
+
+        self.empty_staging();
+        freshly_committed
+    }
+
+    /// Returns sierra for cairo1; returns the only artifact for cairo0.
+    pub fn get_class(
+        &self,
+        class_hash: &ClassHash,
+        block_id: &BlockTagOrNumber,
```
mikiw, line 99: here is block_id but type is BlockTagOrNumber

mikiw (changes requested):

crates/starknet-devnet-core/src/starknet/get_class_impls.rs
```diff
@@ -25,12 +25,23 @@ pub fn get_class_hash_at_impl(
 }
 
 pub fn get_class_impl(
-    starknet: &mut Starknet,
+    starknet: &mut Starknet, // TODO can be regular reference?
```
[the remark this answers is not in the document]
FabijanC: I'm surprised clippy doesn't warn about this

crates/starknet-devnet-core/src/starknet/mod.rs
FabijanC: before this change, we had this logic of finding the starting block number in two places (here and in Starknet::new).

crates/starknet-devnet-core/src/blocks/mod.rs
FabijanC: it's not in case of forking, and it's later needed in block abortion; see https://github.com/0xSpaceShard/starknet-devnet-rs/pull/499#discussion_r1656858900

crates/starknet-devnet-core/src/state/mod.rs
FabijanC: well it's still block id; I don't plan on calling it block_tag_or_number

crates/starknet-devnet-core/src/starknet/get_class_impls.rs
```diff
@@ -25,12 +25,23 @@ pub fn get_class_hash_at_impl(
...
     block_id: &BlockId,
     class_hash: ClassHash,
 ) -> DevnetResult<ContractClass> {
-    let state = starknet.get_mut_state_at(block_id)?;
-    match state.get_rpc_contract_class(&class_hash) {
+    starknet.get_block(block_id)?; // returns error if block at the id doesn't exist
+
+    // if block id is hash, convert to number; the underlying logic only works with that or tag
+    let block_tag_or_number = match block_id {
```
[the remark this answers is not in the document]
FabijanC: It _is_ related to the linked comment:
- We work only with block numbers because of speed.
- We can only determine the block number from block hash if we have access to `Starknet.blocks` object - not accessible inside of `Starknet.rpc_contract_classes`.

crates/starknet-devnet-core/src/starknet/mod.rs
```diff
@@ -328,7 +334,9 @@ impl Starknet {
         });
 
         // insert pending block in the blocks collection and connect it to the state diff
-        self.blocks.insert(new_block, state_diff);
+        self.blocks.insert(new_block, self.pending_state_diff.clone());
```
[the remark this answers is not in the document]
FabijanC: I agree the implementation is bad, but I'm not sure if I know how to make it better. I'm currently investigating.

crates/starknet-devnet-core/src/state/mod.rs
mikiw: well in the whole project when we have variables called `block_id` we refer to BlockId enum, when we use `block_tag` we refer to `BlockTag`, when we use `block_number` we refer to `u64` and when we use 
 `block_hash` we refer to BlockHash
 
 so my expectation here is to have `block_tag_number`, or something like this to be consistent with the whole project

mikiw: but I don't wanna escalate Parkinson's law here since there are more important stuff to do, feel free to resolve it if you want

FabijanC: Addressed in https://github.com/0xSpaceShard/starknet-devnet-rs/pull/499/commits/7b61c28c85491cc5607b69340f2378de2ef5d53c

crates/starknet-devnet-core/src/state/mod.rs
```diff
@@ -55,72 +61,146 @@ pub trait CustomState {
...
         self.staging.insert(class_hash, contract_class);
     }
 
-    pub fn commit(&mut self) -> HashMap<ClassHash, ContractClass> {
-        let diff = self.staging.clone();
-        self.committed.extend(self.staging.drain());
-        diff
+    /// Commits all of the staged classes and returns them, together with their hashes.
+    pub fn commit(&mut self, block_number: u64) -> HashMap<ClassHash, ContractClass> {
+        let mut freshly_committed = HashMap::new();
```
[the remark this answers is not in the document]
FabijanC: Changed name to `newly_committed` in https://github.com/0xSpaceShard/starknet-devnet-rs/pull/499/commits/f63cae06df67aab37672c9dd3aa0ca3967c6962b

crates/starknet-devnet-core/src/starknet/get_class_impls.rs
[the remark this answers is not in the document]
FabijanC: Thanks for reminding about this TODO; addressed it in https://github.com/0xSpaceShard/starknet-devnet-rs/pull/499/commits/94ade42b16f64bc9a39dca250435ef85e06dcf14

crates/starknet-devnet-core/src/starknet/mod.rs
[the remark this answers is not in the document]
mikiw: My expectation here is better documentation in code

crates/starknet-devnet-core/src/starknet/mod.rs
mikiw: ok got it

mikiw: ok got it

crates/starknet-devnet-core/src/starknet/mod.rs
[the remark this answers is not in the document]
FabijanC: Merged `commit_diff` and `commit_classes` in https://github.com/0xSpaceShard/starknet-devnet-rs/pull/499/commits/704aa34d222ba34f69a2a92bbd4f9d47cdc7a547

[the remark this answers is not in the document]
mikiw: cool

mikiw (approved):
I would link it with https://github.com/0xSpaceShard/starknet-devnet-rs/issues/272

PR #531, opened 2024-07-03
Improve dumping UX

### LINKED ISSUE #529: Dumping UX can be improved
## Problem 

Currently if you just want to dump on request, you need to start Devnet with `--dump-on <SOMETHING>` and `--dump-path <PATH>`.

## Proposal

Add `--dump-on request` which doesn't require `--dump-path` to be set.

Additionally, in the `/dump` endpoint, if no path is specified in the request body (and if none is provided on startup with `--dump-path`), default to dumping to stdout. This will allow the common pattern of:

```bash
curl -X POST <DEVNET_URL>/dump > dump.json
```

## Needed changes

Something should probably be done here:
https://github.com/0xSpaceShard/starknet-devnet-rs/blob/1d959080db5520f670ea5253b905c334d0e874e1/crates/starknet-devnet/src/cli.rs#L141-L158

Might be useful:
https://docs.rs/clap/latest/clap/struct.Arg.html#method.requires_if

3alpha:
## Usage related changes

Improves ux when using `/dump` endpoint. When dump path not provided, it dumps to STDOUT.

Resolves #529 

## Checklist:

### DIFF
--- a/crates/starknet-devnet-core/src/starknet/dump.rs
+++ b/crates/starknet-devnet-core/src/starknet/dump.rs
@@ -62,7 +62,7 @@ impl Starknet {
                 self.dump_events.push(event);
 
                 Ok(())
-            },
+            }
             Some(DumpOn::Exit) => {
                 self.dump_events.push(event);
 
--- a/crates/starknet-devnet-core/src/starknet/dump.rs
+++ b/crates/starknet-devnet-core/src/starknet/dump.rs
@@ -58,6 +58,11 @@ impl Starknet {
     pub fn handle_dump_event(&mut self, event: DumpEvent) -> DevnetResult<()> {
         match self.config.dump_on {
             Some(DumpOn::Block) => self.dump_event(event),
+            Some(DumpOn::Request) => {
+                self.dump_events.push(event);
+
+                Ok(())
+            },
             Some(DumpOn::Exit) => {
                 self.dump_events.push(event);
 
@@ -131,6 +136,10 @@ impl Starknet {
             }
             None => Err(Error::FormatError),
         }
+    }
+
+    pub fn dump_events_vec(&self) -> DevnetResult<Vec<DumpEvent>> {
+        Ok(self.dump_events.clone())
     }
 
     pub fn load_events(&self) -> DevnetResult<Vec<DumpEvent>> {
--- a/crates/starknet-devnet-core/src/starknet/dump.rs
+++ b/crates/starknet-devnet-core/src/starknet/dump.rs
@@ -58,12 +58,7 @@ impl Starknet {
     pub fn handle_dump_event(&mut self, event: DumpEvent) -> DevnetResult<()> {
         match self.config.dump_on {
             Some(DumpOn::Block) => self.dump_event(event),
-            Some(DumpOn::Request) => {
-                self.dump_events.push(event);
-
-                Ok(())
-            }
-            Some(DumpOn::Exit) => {
+            Some(DumpOn::Request | DumpOn::Exit) => {
                 self.dump_events.push(event);
 
                 Ok(())
--- a/crates/starknet-devnet-core/src/starknet/dump.rs
+++ b/crates/starknet-devnet-core/src/starknet/dump.rs
@@ -133,8 +133,8 @@ impl Starknet {
         }
     }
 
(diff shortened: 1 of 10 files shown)


crates/starknet-devnet-core/src/starknet/dump.rs
```diff
@@ -133,6 +138,10 @@ impl Starknet {
         }
     }
 
+    pub fn dump_events_vec(&self) -> DevnetResult<Vec<DumpEvent>> {
```
FabijanC, line 141: Let's not introduce Hungarian notation by mentioning the vector. Also, no need for this to return `Result` or `DevnetResult`. In fact, it could return a mere reference:
```rust
 pub fn read_dump_events(&self) -> &Vec<DumpEvent> {
```

crates/starknet-devnet-server/src/api/http/models.rs
```diff
@@ -41,6 +42,8 @@ pub struct MessageHash {
 pub struct TxHash {
     pub transaction_hash: TransactionHash,
 }
+// Implemented as type alias so JSON returned doesn't have extra key
+pub type Dump = Option<Vec<DumpEvent>>;
```
FabijanC, line 46: I find this type misleading: a `Dump` should always be something. A dump response body, on the other hand, could be optional. It would be better for those functions whose return type is `SomeWrapper<Dump>` to rather rather return `SomeWrapper<Option<Vec<DumpEvent>>>`. Or perhaps just rename `Dump` to `DumpResponseBody`.

FabijanC (changes requested):
I'd add an e2e test which creates a tx (usually we just add a dummy mint), requests a dump (which is redirected into a file) and then uses the dump to load a new devnet and check if the state is indeed as it was before dumping (i.e. if the balance of a dummy address is changed by the performed mint).

crates/starknet-devnet-server/src/api/http/endpoints/dump_load.rs
```diff
@@ -25,24 +25,37 @@ pub(crate) async fn dump_impl(api: &Api, path: DumpPath) -> HttpApiResult<()> {
     match path.path {
         None => {
             // path not present
-            starknet
-                .dump_events()
-                .map_err(|err| HttpApiError::DumpError { msg: err.to_string() })?;
-            Ok(())
+            match starknet.config.dump_path.clone() {
```
FabijanC, line 28: Forgot to mention that this line can match a reference, and later the cloning can be done in L32.
```suggestion
            match &starknet.config.dump_path{
```

mikiw (approved):
looks good

crates/starknet-devnet/tests/common/background_devnet.rs
```diff
@@ -362,6 +362,10 @@ impl BackgroundDevnet {
         self.send_custom_rpc("devnet_getConfig", json!({})).await.unwrap()
     }
 
+    pub async fn get_dump(&self) -> serde_json::Value {
+        self.send_custom_rpc("devnet_dump", json!({})).await.unwrap()
+    }
+
```
FabijanC, line 368: We don't need this method if it's used in just one place. Or if we really want to keep it, it could return a `Result<...>`

crates/starknet-devnet/tests/test_dump_and_load.rs
```diff
@@ -57,6 +57,33 @@ mod dump_and_load_tests {
...
+            "--dump-path",
+            &dump_file.path,
+            "--dump-on",
+            "request",
+        ])
+        .await
+        .expect("Could not start Devnet");
+
+        let last_block = devnet_load.get_latest_block_with_tx_hashes().await.unwrap();
+        assert_eq!(last_block.block_number, 4);
```
FabijanC, line 84: Why not also assert balance (the code is not formatted)

```rust
let loaded_balance = devnet_load.get_balance_latest(&FieldElement::from(DUMMY_ADDRESS), FeeUnit::WEI).await.unwrap();
assert_eq!(loaded_balance, FieldElement::from(DUMMY_AMOUNT * 2));
```

FabijanC (changes requested):

crates/starknet-devnet/tests/common/background_devnet.rs
3alpha: Will remove it, just introduced it for consistency.

FabijanC (approved):

PR #548, opened 2024-07-09
Fix nightly version in CI

marioiordanov:

CI pipeline uses fixed version of nightly toolchain. The reason for it is that in the CI nightly channel is downloaded, which is the latest one. There were cases where the devs nightly channel on their computer was different from the downloaded during CI pipeline and it broke the workflow

## Checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -55,7 +55,6 @@ jobs:
           keys:
             - << parameters.cargo_cache_key >>
       - run:
-          # this step doesn't need to be cached, it's relatively fast (15s on Docker/X-Large)
           name: Install nightly-2024-07-08
           command: rustup install nightly-2024-07-08
       - run:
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -79,6 +79,7 @@ jobs:
           # Test also generates, but significantly more
           # Compromise: saving before Test
           paths:
+            - /home/circleci/.rustup/toolchains
```
FabijanC, line 82: If we're going to cache it, the comment in L58 needs to be removed.

FabijanC (changes requested):

.circleci/config.yml
FabijanC: Now that I think about it, once we change the nightly version, it will invalidate the cache because the cache key is not affected by the nightly version:
```
cargo_cache_key: cargo-cache-{{ checksum "Cargo.lock" }}-{{ checksum "rust-toolchain.toml" }}-v1
```

It shouldn't cause problems though, because cargo will automatically download the new nightly if it's not there. Let me know if you have thoughts about this.

FabijanC (approved):

PR #558, opened 2024-07-18
Restrictive mode

marioiordanov:
## Usage related changes

Users can start devnet with cli argument `--restrictive-mode` or `--restrictive-mode <method_1> <method_2>`

## Checklist:

### DIFF
--- a/crates/starknet-devnet-server/src/api/json_rpc/mod.rs
+++ b/crates/starknet-devnet-server/src/api/json_rpc/mod.rs
@@ -124,14 +124,7 @@ impl RpcHandler for JsonRpcHandler {
         match serde_json::from_value::<Self::Request>(deserializable_call) {
             Ok(req) => {
                 if let Some(restricted_methods) = &self.server_config.restricted_methods {
-                    if is_json_rpc_method_restricted(
-                        &method,
-                        restricted_methods
-                            .iter()
-                            .map(|x| x.as_str())
-                            .collect::<Vec<&str>>()
-                            .as_slice(),
-                    ) {
+                    if is_json_rpc_method_restricted(&method, restricted_methods) {
                         return RpcResponse::new(
                             id,
                             RpcError::new(crate::rpc_core::error::ErrorCode::MethodForbidden),
--- a/crates/starknet-devnet-server/src/api/json_rpc/mod.rs
+++ b/crates/starknet-devnet-server/src/api/json_rpc/mod.rs
@@ -8,7 +8,7 @@ mod write_endpoints;
 
 pub const RPC_SPEC_VERSION: &str = "0.7.1";
 
-use enum_helper_macros::AllVariantsSerdeRenames;
+use enum_helper_macros::{AllVariantsSerdeRenames, VariantName};
 use models::{
     BlockAndClassHashInput, BlockAndContractAddressInput, BlockAndIndexInput, CallInput,
     EstimateFeeInput, EventsInput, GetStorageInput, TransactionHashInput, TransactionHashOutput,
@@ -316,7 +316,7 @@ impl JsonRpcHandler {
     }
 }
 
-#[derive(Deserialize, AllVariantsSerdeRenames)]
+#[derive(Deserialize, AllVariantsSerdeRenames, VariantName)]
 #[cfg_attr(test, derive(Debug))]
 #[serde(tag = "method", content = "params")]
 pub enum JsonRpcRequest {
@@ -419,72 +419,9 @@ pub enum JsonRpcRequest {
     #[serde(rename = "devnet_getConfig", with = "empty_params")]
     DevnetConfig,
 }
-
-impl JsonRpcRequest {
-    pub fn method_name(&self) -> &'static str {
-        match self {
-            JsonRpcRequest::SpecVersion => "starknet_specVersion",
-            JsonRpcRequest::BlockWithTransactionHashes(_) => "starknet_getBlockWithTxHashes",
-            JsonRpcRequest::BlockWithFullTransactions(_) => "starknet_getBlockWithTxs",
-            JsonRpcRequest::BlockWithReceipts(_) => "starknet_getBlockWithReceipts",
-            JsonRpcRequest::StateUpdate(_) => "starknet_getStateUpdate",
-            JsonRpcRequest::StorageAt(_) => "starknet_getStorageAt",
-            JsonRpcRequest::TransactionByHash(_) => "starknet_getTransactionByHash",
-            JsonRpcRequest::TransactionStatusByHash(_) => "starknet_getTransactionStatus",
-            JsonRpcRequest::TransactionByBlockAndIndex(_) => {
-                "starknet_getTransactionByBlockIdAndIndex"
-            }
-            JsonRpcRequest::TransactionReceiptByTransactionHash(_) => {
-                "starknet_getTransactionReceipt"
-            }
-            JsonRpcRequest::ClassByHash(_) => "starknet_getClass",
-            JsonRpcRequest::ClassHashAtContractAddress(_) => "starknet_getClassHashAt",
-            JsonRpcRequest::ClassAtContractAddress(_) => "starknet_getClassAt",
-            JsonRpcRequest::BlockTransactionCount(_) => "starknet_getBlockTransactionCount",
-            JsonRpcRequest::Call(_) => "starknet_call",
-            JsonRpcRequest::EstimateFee(_) => "starknet_estimateFee",
-            JsonRpcRequest::BlockNumber => "starknet_blockNumber",
-            JsonRpcRequest::BlockHashAndNumber => "starknet_blockHashAndNumber",
-            JsonRpcRequest::ChainId => "starknet_chainId",
-            JsonRpcRequest::Syncing => "starknet_syncing",
-            JsonRpcRequest::Events(_) => "starknet_getEvents",
-            JsonRpcRequest::ContractNonce(_) => "starknet_getNonce",
-            JsonRpcRequest::AddDeclareTransaction(_) => "starknet_addDeclareTransaction",
-            JsonRpcRequest::AddDeployAccountTransaction(_) => {
-                "starknet_addDeployAccountTransaction"
-            }
-            JsonRpcRequest::AddInvokeTransaction(_) => "starknet_addInvokeTransaction",
-            JsonRpcRequest::EstimateMessageFee(_) => "starknet_estimateMessageFee",
-            JsonRpcRequest::SimulateTransactions(_) => "starknet_simulateTransactions",
-            JsonRpcRequest::TraceTransaction(_) => "starknet_traceTransaction",
-            JsonRpcRequest::BlockTransactionTraces(_) => "starknet_traceBlockTransactions",
-            JsonRpcRequest::ImpersonateAccount(_) => "devnet_impersonateAccount",
-            JsonRpcRequest::StopImpersonateAccount(_) => "devnet_stopImpersonateAccount",
-            JsonRpcRequest::AutoImpersonate => "devnet_autoImpersonate",
-            JsonRpcRequest::StopAutoImpersonate => "devnet_stopAutoImpersonate",
-            JsonRpcRequest::Dump(_) => "devnet_dump",
-            JsonRpcRequest::Load(_) => "devnet_load",
-            JsonRpcRequest::PostmanLoadL1MessagingContract(_) => "devnet_postmanLoad",
-            JsonRpcRequest::PostmanFlush(_) => "devnet_postmanFlush",
-            JsonRpcRequest::PostmanSendMessageToL2(_) => "devnet_postmanSendMessageToL2",
-            JsonRpcRequest::PostmanConsumeMessageFromL2(_) => "devnet_postmanConsumeMessageFromL2",
-            JsonRpcRequest::CreateBlock => "devnet_createBlock",
-            JsonRpcRequest::AbortBlocks(_) => "devnet_abortBlocks",
-            JsonRpcRequest::SetGasPrice(_) => "devnet_setGasPrice",
-            JsonRpcRequest::Restart => "devnet_restart",
-            JsonRpcRequest::SetTime(_) => "devnet_setTime",
-            JsonRpcRequest::IncreaseTime(_) => "devnet_increaseTime",
-            JsonRpcRequest::PredeployedAccounts(_) => "devnet_getPredeployedAccounts",
-            JsonRpcRequest::AccountBalance(_) => "devnet_getAccountBalance",
-            JsonRpcRequest::Mint(_) => "devnet_mint",
-            JsonRpcRequest::DevnetConfig => "devnet_getConfig",
-        }
-    }
-}
-
 impl std::fmt::Display for JsonRpcRequest {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(f, "{}", self.method_name())
+        write!(f, "{}", self.variant_name())
     }
 }
 
--- a/crates/starknet-devnet-server/src/api/json_rpc/mod.rs
+++ b/crates/starknet-devnet-server/src/api/json_rpc/mod.rs
@@ -138,7 +138,6 @@ impl RpcHandler for JsonRpcHandler {
                         );
                     }
                 }
-                // sis_json_rpc_method_restricted(json_rpc_method, self.server_config.)
                 let result = self.on_request(req, call).await;
                 RpcResponse::new(id, result)
             }
--- a/crates/starknet-devnet-server/src/api/json_rpc/mod.rs
+++ b/crates/starknet-devnet-server/src/api/json_rpc/mod.rs
@@ -51,6 +51,7 @@ use crate::api::json_rpc::models::{
     BroadcastedInvokeTransactionEnumWrapper, SimulateTransactionsInput,
 };
 use crate::api::serde_helpers::{empty_params, optional_params};
+use crate::restrictive_methods::is_json_rpc_method_restricted;
 use crate::rpc_core::error::RpcError;
 use crate::rpc_core::request::RpcMethodCall;
 use crate::rpc_core::response::{ResponseResult, RpcResponse};
@@ -122,6 +123,22 @@ impl RpcHandler for JsonRpcHandler {
 
         match serde_json::from_value::<Self::Request>(deserializable_call) {
             Ok(req) => {
+                if let Some(restricted_methods) = &self.server_config.restricted_methods {
+                    if is_json_rpc_method_restricted(
+                        &method,
+                        restricted_methods
+                            .iter()
+                            .map(|x| x.as_str())
+                            .collect::<Vec<&str>>()
+                            .as_slice(),
+                    ) {
+                        return RpcResponse::new(
+                            id,
+                            RpcError::new(crate::rpc_core::error::ErrorCode::MethodForbidden),
+                        );
+                    }
+                }
+                // sis_json_rpc_method_restricted(json_rpc_method, self.server_config.)
                 let result = self.on_request(req, call).await;
                 RpcResponse::new(id, result)
             }
--- a/crates/starknet-devnet-server/src/api/json_rpc/mod.rs
+++ b/crates/starknet-devnet-server/src/api/json_rpc/mod.rs
@@ -53,7 +53,7 @@ use crate::api::json_rpc::models::{
 use crate::api::serde_helpers::{empty_params, optional_params};
 use crate::rpc_core::error::RpcError;
 use crate::rpc_core::request::RpcMethodCall;
-use crate::rpc_core::response::ResponseResult;
+use crate::rpc_core::response::{ResponseResult, RpcResponse};
 use crate::rpc_handler::RpcHandler;
 use crate::ServerConfig;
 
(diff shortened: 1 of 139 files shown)


crates/starknet-devnet/src/cli.rs
```diff
@@ -198,10 +199,13 @@ Sending POST /create_block is also an option in modes other than \"demand\".")]
...
-    #[arg(long = "disable-account-impersonation")]
-    #[arg(env = "DISABLE_ACCOUNT_IMPERSONATION")]
-    #[arg(help = "Disables the possibility to impersonate accounts;")]
-    disable_account_impersonation: bool,
+    #[arg(long = "restrictive-mode")]
+    #[arg(env = "RESTRICTIVE_MODE")]
+    #[arg(num_args = 0..)]
+    #[arg(help = "Use devnet in restrictive mode; You can specify the methods that will be \
+                  fobidden (empty space separated values). If nothing is specified for this \
+                  argument, then default restricted methods are used.")]
```
FabijanC, line 207: - Devnet should be capitalized.
- "fobidden"
- What are "empty space separated values"?
- How will the user know the default restricted methods? Perhaps this could include a link to that paragraph in the docs (we can predict the exact paragraph link, even though it might not yet exist before merging).

crates/starknet-devnet-server/src/restrictive_methods.rs
FabijanC, line 1: I think this is not a correct name for the module. The methods aren't themselves restrictive. We could say Devnet is restrictive regarding the methods, or that the methods are restricted. Perhaps it would make sense to call the module `method_restriction`.

crates/starknet-devnet-server/src/restrictive_methods.rs
```diff
@@ -0,0 +1,119 @@
...
+    static ref MAPPING_RPC_METHODS_TO_HTTP_URIS: HashMap<&'static str, &'static str> =
+        MAPPING_HTTP_URIS_TO_RPC_METHODS.iter().map(|(k, v)| (*v, *k)).collect();
+    pub static ref DEFAULT_RESTRICTED_JSON_RPC_METHODS: Vec<&'static str> = vec![
+        "devnet_mint",
+        "devnet_restart",
+        "devnet_createBlock",
+        "devnet_abortBlocks",
+        "devnet_impersonateAccount",
+        "devnet_autoImpersonate",
+        "devnet_getPredeployedAccounts"
```
FabijanC, line 32: So now when a new method is added which needs to be restricted, what does the developer need to do? Populate this vector, add an entry to the mapping above? We need a list for this, perhaps in CONTRIBUTING.md. The list should actually contain general info regarding new method addition, which should also mention what needs to be done regarding dumping/loading. If you want, we can deal with this in a separate issue, but we must not forget it.

crates/starknet-devnet-server/src/restrictive_methods.rs
```diff
@@ -0,0 +1,119 @@
+use std::collections::HashMap;
+
+use lazy_static::lazy_static;
+lazy_static! {
+    static ref MAPPING_HTTP_URIS_TO_RPC_METHODS: HashMap<&'static str, &'static str> =
+        HashMap::from([
+            ("/dump", "devnet_dump"),
+            ("/load", "devnet_load"),
+            ("/set_time", "devnet_setTime"),
+            ("/increase_time", "devnet_increaseTime"),
```
FabijanC, line 10: This is getting out of hand, each method name is hardcoded in four places in the code.

crates/starknet-devnet-server/src/restrictive_methods.rs
```diff
@@ -0,0 +1,119 @@
...
+            ("/mint", "devnet_mint"),
+            ("/postman/load_l1_messaging_contract", "devnet_postmanLoad"),
+            ("/postman/flush", "devnet_postmanFlush"),
+            ("/postman/send_message_to_l2", "devnet_postmanSendMessageToL2"),
+            ("/postman/consume_message_from_l2", "devnet_postmanConsumeMessageFromL2"),
+            ("/predeployed_accounts", "devnet_getPredeployedAccounts"),
+            ("/account_balance", "devnet_getAccountBalance"),
+            ("/config", "devnet_getConfig"),
+        ]);
+    static ref MAPPING_RPC_METHODS_TO_HTTP_URIS: HashMap<&'static str, &'static str> =
```
FabijanC, line 23: Why this Hungarian notation? The name could be simplifed without loss of information to something like `RPC_METHOD_TO_HTTP_URI` and vice versa.

crates/starknet-devnet-server/src/restrictive_methods.rs
```diff
@@ -0,0 +1,119 @@
...
+    pub static ref DEFAULT_RESTRICTED_JSON_RPC_METHODS: Vec<&'static str> = vec![
+        "devnet_mint",
+        "devnet_restart",
+        "devnet_createBlock",
+        "devnet_abortBlocks",
+        "devnet_impersonateAccount",
+        "devnet_autoImpersonate",
+        "devnet_getPredeployedAccounts"
+    ];
+    static ref DEFAULT_RESTRICTED_HTTP_URIS: Vec<&'static str> =
```
FabijanC, line 34: Isn't this used only in tests?

crates/starknet-devnet-server/src/restrictive_methods.rs
```diff
@@ -0,0 +1,119 @@
...
+
+pub(crate) fn is_uri_path_restricted(uri_path: &str, restricted_uris: &[&str]) -> bool {
+    if restricted_uris.contains(&uri_path) {
+        return true;
+    }
+
+    match MAPPING_HTTP_URIS_TO_RPC_METHODS.get(uri_path) {
+        Some(json_rpc_method) => restricted_uris.contains(json_rpc_method),
+        None => false,
+    }
```
FabijanC, line 64: Something looks wrong in these two methods. Why are we checking if `http_uri` is contained in `restricted_methods`? Shouldn't that always be false?

Is this just to allow users to specify methods as well as uri paths on startup?

crates/starknet-devnet-server/src/rpc_core/error.rs
```diff
@@ -94,6 +94,8 @@ pub enum ErrorCode {
     TransactionRejected,
     /// Custom geth error code, <https://github.com/vapory-legacy/wiki/blob/master/JSON-RPC-Error-Codes-Improvement-Proposal.md>
     ExecutionError,
+    /// method is forbidden for execution
+    MethodForbidden,
```
FabijanC, line 98: Perhaps, from the users' perspective, to make the terminology clearer and more uniform, since we already mention the terms `restrictive` and `restricted`, we could replace `MethodForbidden` with `MethodRestricted` (similar for the message `"Method forbidden"`).

crates/starknet-devnet-server/src/rpc_handler.rs
```diff
@@ -32,36 +32,7 @@ pub trait RpcHandler: Clone + Send + Sync + 'static {
...
-                    error!(target: "rpc", ?method, "failed to deserialize method due to unknown variant");
-                    RpcResponse::new(id, RpcError::method_not_found())
-                } else {
-                    error!(target: "rpc", ?method, ?err, "failed to deserialize method");
-                    RpcResponse::new(id, RpcError::invalid_params(err))
-                }
-            }
-        }
-    }
+    async fn on_call(&mut self, call: RpcMethodCall) -> RpcResponse;
```
FabijanC, line 35: Why was the implementation of this method moved?

crates/starknet-devnet-server/src/server.rs
```diff
@@ -150,3 +158,19 @@ async fn response_logging_middleware(
...
+    State(server_config): State<ServerConfig>,
+    request: Request,
+    next: Next,
+) -> Result<impl IntoResponse, (StatusCode, String)> {
+    if let Some(restricted_paths) = &server_config.restricted_methods {
+        if is_uri_path_restricted(
+            request.uri().path(),
+            restricted_paths.iter().map(|x| x.as_str()).collect::<Vec<&str>>().as_slice(),
+        ) {
+            return Err((StatusCode::FORBIDDEN, "Devnet is in restricted mode".to_string()));
```
FabijanC, line 172: "restricted" or "restrictive"?

crates/starknet-devnet-server/src/restrictive_methods.rs
```diff
@@ -0,0 +1,119 @@
...
+mod tests {
+    use super::{DEFAULT_RESTRICTED_HTTP_URIS, DEFAULT_RESTRICTED_JSON_RPC_METHODS};
+    use crate::restrictive_methods::{is_json_rpc_method_restricted, is_uri_path_restricted};
+
+    #[test]
+    fn test_mappings_length_because_json_rpc_methods_are_greater_than_http_methods() {
+        assert!(
+            crate::restrictive_methods::DEFAULT_RESTRICTED_JSON_RPC_METHODS.len()
+                > crate::restrictive_methods::DEFAULT_RESTRICTED_HTTP_URIS.len()
+        )
```
FabijanC, line 77: Why is this the case? And wouldn't it be better to test some kind of superset/subset relation than plain length comparison?

crates/starknet-devnet-server/src/api/json_rpc/mod.rs
```diff
@@ -109,6 +110,52 @@ impl RpcHandler for JsonRpcHandler {
...
+        match serde_json::from_value::<Self::Request>(deserializable_call) {
+            Ok(req) => {
+                if let Some(restricted_methods) = &self.server_config.restricted_methods {
+                    if is_json_rpc_method_restricted(
+                        &method,
+                        restricted_methods
+                            .iter()
+                            .map(|x| x.as_str())
+                            .collect::<Vec<&str>>()
+                            .as_slice(),
```
FabijanC, line 133: Here and in the HTTP middleware, we will be doing this conversion on every call? Can't it be simplified, e.g. so that the statics hold `String` instead of `&str`?

crates/starknet-devnet/src/cli.rs
```diff
@@ -237,19 +241,26 @@ impl Args {
...
 
         let RequestResponseLogging { log_request, log_response } =
             RequestResponseLogging::from_rust_log_environment_variable();
 
+        let restricted_methods = self.restricted_methods.as_ref().map(|methods| {
+            if methods.is_empty() {
+                DEFAULT_RESTRICTED_JSON_RPC_METHODS.iter().map(|s| s.to_string()).collect()
+            } else {
+                methods.clone()
+            }
```
FabijanC, line 254: So there is no method validation? I.e. if the strings specified are actual method names?

FabijanC (changes requested):

crates/starknet-devnet-server/src/api/json_rpc/mod.rs
```diff
@@ -109,6 +110,52 @@ impl RpcHandler for JsonRpcHandler {
...
+        });
+
+        match serde_json::from_value::<Self::Request>(deserializable_call) {
+            Ok(req) => {
+                if let Some(restricted_methods) = &self.server_config.restricted_methods {
+                    if is_json_rpc_method_restricted(
+                        &method,
+                        restricted_methods
+                            .iter()
+                            .map(|x| x.as_str())
```
mikiw, line 131: `.map(|x| x.as_str())` can be replaced with more readable `.map(String::as_str)` (applies to whole PR)

crates/starknet-devnet-server/src/server.rs
```diff
@@ -150,3 +158,19 @@ async fn response_logging_middleware(
     let response = Response::from_parts(parts, body);
     Ok(response)
 }
+
+async fn restrictive_middleware(
+    State(server_config): State<ServerConfig>,
+    request: Request,
+    next: Next,
+) -> Result<impl IntoResponse, (StatusCode, String)> {
+    if let Some(restricted_paths) = &server_config.restricted_methods {
```
mikiw, line 167: I think that it can be simplified with a match instead of if

crates/starknet-devnet/src/cli.rs
```diff
@@ -237,19 +241,26 @@ impl Args {
                 url: self.fork_network.clone(),
                 block_number: self.fork_block,
             },
-            disable_account_impersonation: self.disable_account_impersonation,
         };
 
         let RequestResponseLogging { log_request, log_response } =
             RequestResponseLogging::from_rust_log_environment_variable();
 
+        let restricted_methods = self.restricted_methods.as_ref().map(|methods| {
```
mikiw, line 249: I think that it can be simplified with a match instead of if

mikiw (changes requested):

crates/starknet-devnet-server/src/api/json_rpc/mod.rs
marioiordanov: https://github.com/0xSpaceShard/starknet-devnet-rs/commit/edf343ce913446fe388c305d2c6d5ea0c668118a

crates/starknet-devnet-server/src/server.rs
marioiordanov: how to make it continue with the default behavior if the method is not from the restricted ones

crates/starknet-devnet-server/src/api/json_rpc/mod.rs
marioiordanov: refactored in https://github.com/0xSpaceShard/starknet-devnet-rs/commit/edf343ce913446fe388c305d2c6d5ea0c668118a

website/docs/restrictive.md
```diff
@@ -0,0 +1,29 @@
+# Restrictive mode
+
+The --restrictive-mode argument enables a restrictive mode for devnet, allowing you to specify methods that are forbidden during execution. This option ensures that certain operations are restricted, enhancing control over the devnet behavior. When user sends a request to one of the restricted methods it will return either JSON-RPC error with code -32604 or HTTP status 403.
+
+## Usage
+
+Devnet will restrict default methods:
+
+```
+$ starknet-devnet --restrictive-mode
```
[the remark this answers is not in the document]
FabijanC: makes sense, let's keep `--restrictive-mode` then

website/docs/restrictive.md
```diff
@@ -0,0 +1,29 @@
...
+
+When no methods are specified, the following default methods will be restricted and their HTTP endpoints counterparts (if any):
+
+- devnet_mint
+- devnet_restart
+- devnet_createBlock
+- devnet_abortBlocks
+- devnet_impersonateAccount
+- devnet_autoImpersonate
+- devnet_getPredeployedAccounts
```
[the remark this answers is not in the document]
FabijanC: It doesn't just prevent tweaking the CTF, but also prevens users from harmfully impacting the state (deliberately or not).

crates/starknet-devnet-server/src/restrictive_methods.rs
FabijanC: I'm also ok with `restrictive_mode` as the file name

crates/starknet-devnet/src/cli.rs
FabijanC: I thought you meant that, but I think that "empty space separated" is more commonly expressed as "whitespace-separated" (with the hyphen)

crates/starknet-devnet-server/src/server.rs
FabijanC: Since @mikiw is not likely to respond here, I'll resolve this and the other related comment.

crates/starknet-devnet-server/src/restrictive_methods.rs
marioiordanov: https://github.com/0xSpaceShard/starknet-devnet-rs/commit/17b09cbe5d96dc6a88f42f058c5c34e17f4f7f35

FabijanC: IMO the problem is not so much in hardcoding vs using constants. We need a solution that will allow us to specify the method name in a single place, but that's out of scope for this PR. Resolving.

crates/starknet-devnet-server/src/restrictive_methods.rs
FabijanC: So as suggested in the last question of my previous comment, it is to allow specifying methods and URI paths. But it still looks wrong; it means that a container called `restricted_uris` contains rpc method names AND http uris.

crates/starknet-devnet/src/cli.rs
FabijanC: So if a user specifies `--restrictive-mode mintt dump load`, minting will still be possible, without the user knowing they made a mistake?

crates/starknet-devnet-server/src/restrictive_methods.rs
marioiordanov: yes, its possible. User specifies whatever he wants.

crates/starknet-devnet/src/cli.rs
marioiordanov: yes

marioiordanov: https://github.com/0xSpaceShard/starknet-devnet-rs/commit/bd705fa24b8e143696bcc3a47149f48f610956be..11e8f9c22b556b9f2e2829d2fc3841bef30d9990
https://github.com/0xSpaceShard/starknet-devnet-rs/commit/c857060ecc91bdeff62ef0b99b88ca4bed8147fc
https://github.com/0xSpaceShard/starknet-devnet-rs/commit/adcb70e1ed27c75899435af2d9588d44a40e19d2
https://github.com/0xSpaceShard/starknet-devnet-rs/commit/0dd445aa576f6a2cea3fca83d84e196f0950ed0b
https://github.com/0xSpaceShard/starknet-devnet-rs/commit/11e8f9c22b556b9f2e2829d2fc3841bef30d9990

crates/starknet-devnet/src/cli.rs
```diff
@@ -652,4 +695,20 @@ mod tests {
             ),
         }
     }
+
+    #[test]
+    fn check_if_method_with_incorrect_name_will_produce_an_error() {
```
FabijanC, line 700: Why not add another test which lists some http paths or rpc methods, but succeeds?

crates/starknet-devnet/src/cli.rs
```diff
@@ -237,19 +244,55 @@ impl Args {
...
+            for method in methods {
+                if !all_methods.contains(method) {
+                    wrong_restricted_methods.push(method.clone());
+                }
+            }
+            if !wrong_restricted_methods.is_empty() {
+                anyhow::bail!(
+                    "Restricted methods contain JSON-RPC methods and/or HTTP routes that are not \
+                     supported by the server: {}",
+                    wrong_restricted_methods.join(" ")
```
FabijanC, line 283: Why not comma-separate them? I.e. `.join(", ")`

crates/starknet-devnet/src/cli.rs
```diff
@@ -237,19 +244,55 @@ impl Args {
...
+            if methods.is_empty() {
+                DEFAULT_RESTRICTED_JSON_RPC_METHODS
+                    .iter()
+                    .map(|s| s.to_string())
+                    .collect::<Vec<String>>()
+            } else {
+                // remove leading slashes
+                methods
+                    .iter()
+                    .map(|s| s.trim_start_matches('/').to_string())
```
FabijanC, line 263: In multiple places we are removing this leading slash, isn't a single place enough?

FabijanC (changes requested):

crates/starknet-devnet/tests/test_restrictive_mode.rs
```diff
@@ -0,0 +1,91 @@
...
+                }),
+            )
+            .await
+            .unwrap_err();
+
+        assert_eq!(json_rpc_error.code, ErrorCode::MethodForbidden);
+    }
+
+    #[tokio::test]
+    async fn restrictive_mode_with_custom_methods() {
```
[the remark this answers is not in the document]
marioiordanov: there is already a test for it: https://github.com/0xSpaceShard/starknet-devnet-rs/blob/f45972052c0ff215983839e9802874ffadb45ae2/crates/starknet-devnet/tests/test_restrictive_mode.rs#L14

crates/starknet-devnet/src/cli.rs
marioiordanov: because they are space separated when specified. Keeping the same format

crates/starknet-devnet/src/cli.rs
marioiordanov: no, because 1 place is modyfing the user input, the other is removing the slash from the uri path of the current request. The other is used to construct a collection of default restricted methods

crates/starknet-devnet/src/cli.rs
marioiordanov: https://github.com/0xSpaceShard/starknet-devnet-rs/commit/34cb8ef811967625e9872d67eceba2179c627b65

crates/starknet-devnet/src/cli.rs
FabijanC: makes sense

crates/starknet-devnet/src/cli.rs
FabijanC: There is also an occurrence in `is_uri_path_restricted`, but maybe it's a good thing.

marioiordanov: it is to check if the uri path is restricted

FabijanC: Ah you mean in the middleware, didn't realize it's used there. Resolving.

PR #559, opened 2024-07-18
Add CI cache on self-hosted runner 

3alpha:
## Usage related changes
 
N/A

Reduces CI time by caching build dependencies locally in our self-hosted runners.

## Checklist:


.circleci/config.yml
```diff
@@ -50,6 +50,9 @@ jobs:
     resource_class: spaceshard/ax41
```
FabijanC, line 50: In L47 and L162 (can't comment on them in this PR, so I'm leaving a note on the closest line), there are still mentions of `cargo_cache_key`. Either we can still rely on that value instead of calculating it ourselves in the new script, or we can completely remove it.

.circleci/scripts/cache.sh
```diff
@@ -0,0 +1,59 @@
...
+        fi
+        tar xvf "$cache_file"
+        ;;
+    "save")
+        echo "Creating cache..."
+        if [ -f "$cache_file" ]; then
+            echo "Cache already exists."
+            exit 0
+        fi
+        tar czvf "$cache_file".tmp ${cached_dirs[@]} # Create a temporary cache file for atomicity
```
FabijanC, line 46: Can you explain the atomicity?

.circleci/scripts/cache.sh
```diff
@@ -0,0 +1,59 @@
...
+fi
+
+action="$1"
+
+# VARIABLES
+# Files to use as cache key
+cache_key_files=("Cargo.lock" "rust-toolchain.toml") 
+
+ # Directories to cache
+cached_dirs=("target/release/.fingerprint target/release/build target/release/deps")
```
FabijanC, line 22: We used to cache paths that are outside of the repository. Also, we cached compilation artifacts from both `release` and `debug` mode:
https://github.com/0xSpaceShard/starknet-devnet-rs/commit/01fb56e24c8b86f6cb8e306cd6221e7a5c415f18#diff-78a8a19706dbd2a4425dd72bdab0502ed7a2cef16365ab7030a5a0588927bf47L80-L88

Why is this no longer necessary?

FabijanC (changes requested):

.circleci/config.yml
3alpha: Yeah, those should be removed completely as they are used only by CircleCI's builtin cache.

.circleci/scripts/cache.sh
3alpha: Maybe atomicity is not the most accurate description, this is to handle race conditions.`*tar.gz` file is created at the start of `tar x` command and it's being changed while `tar` is running. Since we have multiple runners which share this same cache, without this step, there is a possibility that one runner will try to read not completely written cache.

.circleci/scripts/cache.sh
3alpha: - On this runner, global cargo and rustup files are preserved in between runs (toolchains for example)
 - If you look closely, no step before "save cache" does produce debug artifacts so there is no need to save them

FabijanC: > * On this runner, global cargo and rustup files are preserved in between runs (toolchains for example)

Makes sense

> * If you look closely, no step before "save cache" does produce debug artifacts so there is no need to save them

I am aware of that, my question is also: "Should we also cache debug artifacts or not"?

.circleci/scripts/cache.sh
FabijanC: Right, I didn't think about concurrent runs at all. You can add this explantation to the script for future readers. Is there a possibility that two runners will try writing at the same time, even into the tmp file? Is that a problem?

.circleci/config.yml
FabijanC: Removed in https://github.com/0xSpaceShard/starknet-devnet-rs/pull/559/commits/6ab3dbaba77c2370e206848f92d146b2d3e75337
Resolving.

.circleci/scripts/cache.sh
3alpha: Ah, sorry. No clue, that is something that I considered out of scope for this PR and I feel that there is somebody else more qualified than me to make informed decision. I just can note that debug files are pretty big so I am not sure of a time advantage that could be gained from caching as compression and decompression of those files might last nontrivial amount of time.

.circleci/scripts/cache.sh
3alpha: Hm, that is a good point. Not sure, but better safe than sorry. I don't know how I missed half of the scenario. Fixing.

FabijanC: Resolving converstaion after commit https://github.com/0xSpaceShard/starknet-devnet-rs/pull/559/commits/580cbb636b8fa5ef19b659a5ab7f3a4287413cec

FabijanC (approved):

Issue #567, opened 2024-07-24
Several `expect()` calls in production code


FabijanC:
Ignoring tests, there are about ten places in the code where we use `.expect()`, even though a rule we've been following is to always resolve `Result` by means of error propagation and mapping.

To see what I mean, just search the project for `.expect(` with `*test*` files ignored (and filtering out a few other tests and testing utilites not defined in `*test*` files).

The errors expected in `main.rs` are mostly the ones happening on startup, where this kind of handling is IMO more acceptable than in other parts of the code.

Pros of using `.expect()`:
- stack trace available right away, easier debugging

Pros of error propagation:
- More pleasant UX when the error message finally reaches the user.

FabijanC:
Closed via #573

PR #573, opened 2024-08-07
Refactoring of unwraps and expects

marioiordanov:
## Usage related changes

None

closing #567 
Remove some expects in the code, or refactor them to use match statement
Added linters to ./scripts/clippy_check.sh
## Checklist:

### DIFF
--- a/crates/starknet-devnet-core/src/messaging/mod.rs
+++ b/crates/starknet-devnet-core/src/messaging/mod.rs
@@ -82,20 +82,20 @@ impl MessagingBroker {
 
     /// Returns a reference to the ethereum instance if configured, an error otherwise.
     pub fn ethereum_ref(&self) -> DevnetResult<&EthereumMessaging> {
-        if self.ethereum.is_none() {
-            return Err(Error::MessagingError(MessagingError::NotConfigured));
+        if let Some(ethereum) = self.ethereum.as_ref() {
+            Ok(ethereum)
+        } else {
+            Err(Error::MessagingError(MessagingError::NotConfigured))
         }
-
-        Ok(self.ethereum.as_ref().unwrap())
     }
 
     /// Returns a mutable reference to the ethereum instance if configured, an error otherwise.
     pub fn ethereum_mut(&mut self) -> DevnetResult<&mut EthereumMessaging> {
-        if self.ethereum.is_none() {
-            return Err(Error::MessagingError(MessagingError::NotConfigured));
+        if let Some(ethereum) = self.ethereum.as_mut() {
+            Ok(ethereum)
+        } else {
+            Err(Error::MessagingError(MessagingError::NotConfigured))
         }
-
-        Ok(self.ethereum.as_mut().unwrap())
     }
 }
 
--- a/crates/starknet-devnet-core/src/messaging/mod.rs
+++ b/crates/starknet-devnet-core/src/messaging/mod.rs
@@ -82,20 +82,12 @@ impl MessagingBroker {
 
     /// Returns a reference to the ethereum instance if configured, an error otherwise.
     pub fn ethereum_ref(&self) -> DevnetResult<&EthereumMessaging> {
-        if let Some(ethereum) = self.ethereum.as_ref() {
-            Ok(ethereum)
-        } else {
-            Err(Error::MessagingError(MessagingError::NotConfigured))
-        }
+        self.ethereum.as_ref().ok_or(Error::MessagingError(MessagingError::NotConfigured))
     }
 
     /// Returns a mutable reference to the ethereum instance if configured, an error otherwise.
     pub fn ethereum_mut(&mut self) -> DevnetResult<&mut EthereumMessaging> {
-        if let Some(ethereum) = self.ethereum.as_mut() {
-            Ok(ethereum)
-        } else {
-            Err(Error::MessagingError(MessagingError::NotConfigured))
-        }
+        self.ethereum.as_mut().ok_or(Error::MessagingError(MessagingError::NotConfigured))
     }
 }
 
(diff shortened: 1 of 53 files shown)


crates/starknet-devnet-core/src/messaging/mod.rs
```diff
@@ -82,20 +82,20 @@ impl MessagingBroker {
...
     }
 
     /// Returns a mutable reference to the ethereum instance if configured, an error otherwise.
     pub fn ethereum_mut(&mut self) -> DevnetResult<&mut EthereumMessaging> {
-        if self.ethereum.is_none() {
-            return Err(Error::MessagingError(MessagingError::NotConfigured));
+        if let Some(ethereum) = self.ethereum.as_mut() {
+            Ok(ethereum)
+        } else {
+            Err(Error::MessagingError(MessagingError::NotConfigured))
```
FabijanC, line 97: Can be simplified as:
```suggestion
            self.ethereum.as_mut().ok_or(Error::MessagingError(MessagingError::NotConfigured))
```

crates/starknet-devnet-core/src/messaging/mod.rs
```diff
@@ -82,20 +82,20 @@ impl MessagingBroker {
 
     /// Returns a reference to the ethereum instance if configured, an error otherwise.
     pub fn ethereum_ref(&self) -> DevnetResult<&EthereumMessaging> {
-        if self.ethereum.is_none() {
-            return Err(Error::MessagingError(MessagingError::NotConfigured));
+        if let Some(ethereum) = self.ethereum.as_ref() {
+            Ok(ethereum)
+        } else {
+            Err(Error::MessagingError(MessagingError::NotConfigured))
```
FabijanC, line 88: Can be simplified as:
```suggestion
            self.ethereum.as_ref().ok_or(Error::MessagingError(MessagingError::NotConfigured))
```

crates/starknet-devnet-core/src/starknet/mod.rs
```diff
@@ -767,7 +767,8 @@ impl Starknet {
         let raw_execution = RawExecution {
             calls: vec![Call {
                 to: erc20_address.into(),
-                selector: get_selector_from_name("mint").unwrap(),
+                selector: get_selector_from_name("mint")
+                    .map_err(|err| Error::UnexpectedInternalError { msg: err.to_string() })?,
```
FabijanC, line 771: Why in same places we note that it is safe to use unwrap and in others, like this one, that it isn't?

crates/starknet-devnet-core/src/state/mod.rs
```diff
@@ -153,7 +153,7 @@ pub struct StarknetState {
     rpc_contract_classes: Arc<RwLock<CommittedClassStorage>>,
     /// - initially `None`
     /// - indicates the state hasn't yet been cloned for old-state preservation purpose
-    historic_state: Option<DictState>,
+    historic_state: DictState,
```
FabijanC, line 156: I don't remember exactly but think it was important that this was an `Option`. If indeed it is not needed anymore, the `/// documentation` need to be changed.

crates/starknet-devnet/tests/common/mod.rs
```diff
@@ -1,3 +1,4 @@
+#![cfg(test)]
```
FabijanC, line 1: What does the `!` mean?

FabijanC (changes requested):

crates/starknet-devnet/tests/common/mod.rs
marioiordanov: it is to point that it is for the entire file

FabijanC: So before this annotation placed in these files, all of these code parts were compiled into the `--release` binary?

marioiordanov: not, but without this attribute clippy complains about using unwraps. Its a workaround suggested in rust forums

crates/starknet-devnet-core/src/starknet/mod.rs
marioiordanov: There are 2 places with `#[allow(clippy::unwrap/expect)]` the others are in the main file which it is ok if it fails, because it will terminate the program, which is desired effect and in some exported test methods, which to be used in another crates have to be included in the --release build. 
So what is left is StarknetConfig::default method and `get_unix_timestamp_as_seconds`. In `StarknetConfig` if `#[allow]` attribute is removed, then we will lose the trait implementation, because there is no default trait that returns Result.
`get_unix_timestamp_as_seconds` wont fail in the coming 584942417354 years.

crates/starknet-devnet-core/src/state/mod.rs
FabijanC: I think the remaining documentation line was referring to the deleted one.

crates/starknet-devnet-core/src/starknet/mod.rs
FabijanC: > get_unix_timestamp_as_seconds wont fail in the coming 584942417354 years.

Why not be ready when that moment comes? Perhaps we could create an issue so that we don't forget.

FabijanC: Joking. Resolving.

crates/starknet-devnet-core/src/state/mod.rs
marioiordanov: https://github.com/0xSpaceShard/starknet-devnet-rs/commit/6b52603d2225e828b072cb54639d14f3d45683ec

FabijanC (approved):

Issue #613, opened 2024-09-23
JSON-RPC API 0.8.0 adaptation


FabijanC:
- Documentation: 
https://github.com/starkware-libs/starknet-specs/releases/tag/v0.8.0-rc1
- Currently in RC phase: 0.8.0-rc.1
- Subtask PRs should target branch `json-rpc-v0.8.0`
- Requires an adapted Blockifier version (or at least a commit)

## Subtasks

- [x] Error improvements - introduced in #627 
  - failure reason in `getTransactionStatus`
  - structure in execution errors
  - `INSUFFICIENT_MAX_FEE` -> `INSUFFICIENT_RESOURCES_FOR_VALIDATE`
- [ ] Gas consumption changes
- [x] WebSocket API support - introduced in #619
- RPC method additions:
  - [x] `starknet_getCompiledCasm` - introduced in #633
  - [x] `starknet_getStorageProof` - introduced in #659
  - [x] `starknet_getMessagesStatus` - introduced in #621
  - [x] 10 ws-related methods - introduced in several PRs

PR #634, opened 2024-10-25
Add WebSocket RPC methods: subscribeNewHeads, unsubscribe

FabijanC:
## Usage related changes

- Introduce support for `starknet_subscribeNewHeads` and `starknet_unsubscribe`.
  - Other methods using `todo!`.
- Partially address #613 
- Address TODO in `main.rs` which prevented SIGINT from killing Devnet instantly in block-on-interval mode.
- Remove general RPC support, i.e. methods like `starknet_getNonce` or `devnet_mint` are no longer supported via WebSocket. Apparently they never should have been, only `starknet_(un)subscribe` methods should be available via this protocol.
  - Not a breaking change since this was never a part of the `main` branch.

- Socket ID and subscription ID generated randomly - should have low collision probability since they use `i64`.
  - Just search for `rand::random()` to find occurrences of this.
  - Alternatively we could generate in loop until an unused value is generated.
- Refactor `JsonRpcHandler`:
  - Introduce new helper function `to_json_rpc_request`
  - Add new helper method `allows_method` which checks restrictedness, relying on previously defined `is_json_rpc_method_restricted`
- Expand the server's `api` property with  a `sockets` map.
- Parts of code related to unimplemented subscription methods is commented out. This is to prevent warnings suggesting `Box`ing of enum variants.
- Since general RPC support via WebSocket is removed, the corresponding tests are ignored, but a new test is introduced which asserts that the general RPC support is indeed not supported.
- The case for invalid subscription request due to `TooManyBlocksBack` is not tested because I couldn't make it generate 1024 blocks efficiently.

## Checklist:


crates/starknet-devnet-server/src/api/json_rpc/endpoints_ws.rs
```diff
@@ -0,0 +1,111 @@
...
+
+        let starknet = self.api.starknet.lock().await;
+
+        // checking the block's existence; aborted blocks treated as not found
+        let query_block = match starknet.get_block(&block_id) {
+            Ok(block) => match block.status() {
+                BlockStatus::Rejected => Err(ApiError::BlockNotFound),
+                _ => Ok(block),
+            },
+            Err(Error::NoBlock) => Err(ApiError::BlockNotFound),
```
marioiordanov, line 67: what happens if block_id is some future block like 2 blocks after the latest.

crates/starknet-devnet-server/src/subscribe.rs
```diff
@@ -0,0 +1,159 @@
...
+}
+
+#[derive(Debug, Clone)]
+pub enum SubscriptionResponse {
+    Confirmation { rpc_request_id: Id, result: SubscriptionConfirmation },
+    Notification { subscription_id: SubscriptionId, data: Box<SubscriptionNotification> },
+}
+
+impl SubscriptionResponse {
+    fn to_serialized_rpc_response(&self) -> serde_json::Value {
```
marioiordanov, line 67: why not already created structs RpcNotification and RpcResponse/JsonRpcResponse

crates/starknet-devnet-server/src/api/json_rpc/endpoints_ws.rs
```diff
@@ -0,0 +1,111 @@
...
+            Err(other) => Err(ApiError::StarknetDevnetError(other)),
+        }?;
+
+        let latest_block = starknet.get_block(&latest_tag)?;
+
+        let query_block_number = query_block.block_number().0;
+        let latest_block_number = latest_block.block_number().0;
+
+        let blocks_back_amount = if query_block_number > latest_block_number {
```
marioiordanov, line 76: if user cant specify block number in the future, then this if clause is not useful. `query_block_number` will be at least equal to `latest_block_number`

crates/starknet-devnet-server/src/api/json_rpc/mod.rs
```diff
@@ -205,6 +206,37 @@ impl JsonRpcHandler {
         }
     }
 
+    async fn get_latest_block(&self) -> StarknetBlock {
+        let starknet = self.api.starknet.lock().await;
+        match starknet.get_block(&BlockId::Tag(BlockTag::Latest)) {
+            Ok(block) => block.clone(),
+            Err(_) => StarknetBlock::create_empty_accepted(),
```
marioiordanov, line 213: why `create_empty_accepted` is used, isnt it better to return an Option from `get_latest_block`. Otherwise notify_subscribers will push some default data to the socket

crates/starknet-devnet-server/src/api/json_rpc/endpoints_ws.rs
FabijanC: Asked Alex, he [replied last week](https://spaceshard.slack.com/archives/C03QN20522D/p1730113533797169?thread_ts=1729779693.200149&cid=C03QN20522D). It's undocumented behavior currently, but from how I understood his response, a new error (invalid block id) will be added to the spec. So the question is whether we already handle this case or when the spec is updated.

crates/starknet-devnet-server/src/api/json_rpc/endpoints_ws.rs
FabijanC: Added this check to prevent overflow (unsigned integer subtraction can't produce a negative result)

crates/starknet-devnet-server/src/api/json_rpc/mod.rs
FabijanC: I agree it's not ideal, but either returning a Result or an Option requires later handling of Err or None case in two places, which I thought was just unnecessary, especially since the latest block is always defined, so the Err/None case will never happen.

crates/starknet-devnet/src/main.rs
```diff
@@ -348,21 +349,21 @@ async fn create_block_interval(
...
-                let mut dumpable_events = api.dumpable_events.lock().await;
-                info!("Generating block on time interval");
-
-                // manually add event for dumping; alternative: create a client and send request
-                starknet.create_block().map_err(|e| std::io::Error::new(std::io::ErrorKind::Other, e))?;
-                dumpable_events.push(DumpEvent { jsonrpc: Version::V2, method: "devnet_createBlock".into(), params: RequestParams::None, id: Id::Number(0) });
+                // By sending a request, we take care of: 1) dumping 2) notifying subscribers
+                match devnet_client.post(&devnet_address).json(&block_req_body).send().await {
+                    Ok(_) => info!("Generating block on time interval"),
+                    Err(e) => error!("Failed block creation on time interval: {e:?}")
```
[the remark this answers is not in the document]
FabijanC: True. But what is the difference? The propagated error doesn't terminate the program, right? And this is not happening somewhere deep in the code, but just one layer below `main`, and is even defined in `main.rs`.

crates/starknet-devnet-server/src/api/json_rpc/mod.rs
```diff
@@ -380,26 +419,58 @@ impl JsonRpcHandler {
...
+    /// errors is left to individual RPC method handlers and this method returns an empty successful
+    /// Result.
+    async fn on_websocket_rpc_call(
+        &self,
+        call: &RpcMethodCall,
+        socket_id: SocketId,
+    ) -> Result<(), RpcError> {
+        trace!(target: "rpc",  id = ?call.id , method = ?call.method, "received websocket call");
+
+        if !self.allows_method(&call.method) {
```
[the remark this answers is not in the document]
marioiordanov: if allows_method is indeed needed to be used, then the enum will have to be decorated with similar traits as JsonRpcRequest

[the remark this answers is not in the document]
FabijanC: With general RPC support removed, calling `allows_method` is no longer needed. If we restore the general RPC support though... Anyway, I removed it in an incoming commit, haven't pushed yet.

crates/starknet-devnet-server/src/api/json_rpc/endpoints_ws.rs
marioiordanov: is it possible to have overflow when `query_block_number` can't be greater than `latest_block_number` because of this comment: https://github.com/0xSpaceShard/starknet-devnet-rs/pull/634#discussion_r1832657640

FabijanC: It depends on whether we want to define a custom block invalid error now or wait for the spec to be updated.

crates/starknet-devnet-server/src/api/json_rpc/mod.rs
FabijanC: Added comment in https://github.com/0xSpaceShard/starknet-devnet-rs/pull/634/commits/9369969e6cc9fd5a47b00585b4790216e87b9479

crates/starknet-devnet-server/src/api/json_rpc/mod.rs
```diff
@@ -380,26 +419,58 @@ impl JsonRpcHandler {
...
+    async fn on_websocket_call(
+        &self,
+        bytes: &[u8],
+        ws: Arc<Mutex<SplitSink<WebSocket, Message>>>,
+        socket_id: SocketId,
+    ) {
+        let error_serialized = match serde_json::from_slice(bytes) {
+            Ok(rpc_call) => match self.on_websocket_rpc_call(&rpc_call, socket_id).await {
+                Ok(_) => return,
+                Err(e) => serde_json::json!({
```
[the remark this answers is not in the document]
FabijanC: Refactored in https://github.com/0xSpaceShard/starknet-devnet-rs/pull/634/commits/10c19fd9a694b69f746e8c19d99a7e242f88439b

crates/starknet-devnet-server/src/api/json_rpc/mod.rs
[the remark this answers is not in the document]
FabijanC: Removed in https://github.com/0xSpaceShard/starknet-devnet-rs/pull/634/commits/7c7c663c3f4bf1b48d3f9b2b2adc7bd161a6fb38

crates/starknet-devnet/src/main.rs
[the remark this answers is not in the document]
marioiordanov: yes, resolving

crates/starknet-devnet-server/src/subscribe.rs
FabijanC: I noticed them while writing the code, just thought that the structures I wrote give better control over the situation.

marioiordanov: constructing the json object can be delegated to already created types. SubscriptionResponse will stay. Similar to JsonRpcResponse and DevnetResponse

marioiordanov: I don't like both approaches but can't figure out a workaround. Resolving

PR #672, opened 2025-01-09
Fix port discovery loop to address flaky tests. Eliminate zombie processes.

### LINKED ISSUE #673: Flaky tests after #664
After #664, some tests occasionally fail, both locally and in CI. In CI, it's just one or two tests that fail, and they might even succeed on re-run. The definition of flaky tests. Locally, it's dozens of failing tests.

The most common failure is:
```
DevnetNotStartable("No such file or directory (os error 2)")
```

FabijanC:
## Usage related changes

Hopefully, none.

- Close #673
- In `BackgroundDevnet::drop`: `wait` after `kill`.
- In `get_acquired_port`: prevent exiting if ports not available; only exit if max retries reached.
  - This is the thing that most likely fixed the flakiness.
- Since the tests sometimes fail with `DevnetNotStartable("No such file or directory (os error 2)")`, one idea was that the subprocess is somehow not always spawned from the same working directory. So the paths of Devnet's Cargo manifest and the executable binary are no longer relative, but absolute, and extracted into constants.
- Add `RUST_BACKTRACE=full` to the CI testing command.
- Improve error info returned by `BackgroundDevnet`.
- Mutex no longer used in `BackgroundDevnet` spawning - relying on `--port 0` is sufficient to avoid collision
  - Introduces more speed-up - tests passing in 8.5 minutes.
- Introduce `SafeChild` to address early exit of BackgroundDevnet not caught by `Drop`.
  - Fixes the zombie process issue.

## Checklist:

### DIFF
--- a/tests/integration/common/background_devnet.rs
+++ b/tests/integration/common/background_devnet.rs
@@ -153,13 +153,13 @@ impl BackgroundDevnet {
                 .args(Self::add_default_args(args))
                 .stdout(Stdio::piped()) // comment this out for complete devnet stdout
                 .spawn()
-                .map_err(|e| TestError::DevnetNotStartable(e.to_string()))?;
+                .map_err(|e| TestError::DevnetNotStartable(format!("Spawning error: {e:?}")))?;
 
         let sleep_time = time::Duration::from_millis(500);
         let max_retries = 40;
         let port = get_acquired_port(process.id(), sleep_time, max_retries)
             .await
-            .map_err(|e| TestError::DevnetNotStartable(e.to_string()))?;
+            .map_err(|e| TestError::DevnetNotStartable(format!("Cannot determine port: {e:?}")))?;
 
         // now we know the port; check if it can be used to poll Devnet's endpoint
         let client = Client::new();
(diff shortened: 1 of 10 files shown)


tests/integration/common/background_devnet.rs
```diff
@@ -160,28 +148,26 @@ impl BackgroundDevnet {
...
                 .spawn()
-                .map_err(|e| TestError::DevnetNotStartable(e.to_string()))?;
+                .map_err(|e| TestError::DevnetNotStartable(format!("Spawning error: {e:?}")))?;
 
         let sleep_time = time::Duration::from_millis(500);
         let max_retries = 40;
         let port = get_acquired_port(process.id(), sleep_time, max_retries)
             .await
-            .map_err(|e| TestError::DevnetNotStartable(e.to_string()))?;
+            .map_err(|e| TestError::DevnetNotStartable(format!("Cannot determine port: {e:?}")))?;
```
marioiordanov, line 162: what if get_acquired_port returns an error, then the process will not be killed. The same applies to wait_for_successful_response

FabijanC: I was going to say that drop is implicitly called, but just realized this is not the case since the BackgroundDevnet object is generated only at the end of the spawn method. And this has to be the cause of all the zombie processes.

marioiordanov: This is from rust docs: 

> There is no implementation of [Drop](https://doc.rust-lang.org/std/ops/trait.Drop.html) for child processes, so if you do not ensure the Child has exited then it will continue to run, even after the Child handle to the child process has gone out of scope.

FabijanC: Fixed in https://github.com/0xSpaceShard/starknet-devnet-rs/pull/672/commits/08c972b703c4e40785ebc02cc3ebb1b524faf771

FabijanC: Actually fixed here: https://github.com/0xSpaceShard/starknet-devnet-rs/pull/672/commits/e732a7eab58a7bd89bf5daa30d0ca14519a420e7

marioiordanov (approved):

Issue #687, opened 2025-01-20
Tests


FabijanC:
Timestamp tests occasionally fail. Restarting the workflow helps.

Tests known to fail (not sure if there's more):
- test_advancing_time::timestamp_syscall_set_in_future
- test_advancing_time::timestamp_syscall_contract_constructor
  - the most frequently failing timestamp test
  - [CI log](https://app.circleci.com/pipelines/github/0xSpaceShard/starknet-devnet-rs/3034/workflows/429f4275-a8dd-49e0-beb7-e18df9bf5b60/jobs/5018)

Basically, waiting/sleeping for 1 second doesn't increment the time used by syscalls in Starknet contracts; the time stays the same.

Tried extending to sleeping for 1.1 s: https://github.com/0xSpaceShard/starknet-devnet-rs/pull/681

Tried extending to sleeping for 1.5 s: https://github.com/0xSpaceShard/starknet-devnet-rs/commit/8f74b339a2a823192801dc444e0441d5730ddd16

PR #729, opened 2025-03-26
Explicitly wait for new timestamp for time advancing tests

3alpha:
## Usage related changes

None

Tries to resolve flakiness of time advancing tests, specified at #687.

## Checklist:

- [x] Linked the [issues](https://github.com/0xSpaceShard/starknet-devnet/issues) resolvable by this PR - [linking info](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword)
- [x] Updated the tests if needed; all passing - [execution info](https://github.com/0xSpaceShard/starknet-devnet/blob/main/.github/CONTRIBUTING.md#test-execution)

### DIFF
--- a/tests/integration/test_advancing_time.rs
+++ b/tests/integration/test_advancing_time.rs
@@ -21,8 +21,12 @@ const DUMMY_AMOUNT: u128 = 1;
 const BUFFER_TIME_SECONDS: u64 = 30;
 
 async fn sleep_until_new_timestamp() {
-    // Sometimes sleeping for 1 second isn't enough.
-    tokio::time::sleep(time::Duration::from_millis(1500)).await
+    let now = get_unix_timestamp_as_seconds();
+    tokio::time::sleep(time::Duration::from_millis(100)).await;
+
+    while now == get_unix_timestamp_as_seconds() {
+        tokio::time::sleep(time::Duration::from_millis(100)).await;
+    }
 }
 
 /// Set time and generate a new block


tests/integration/test_advancing_time.rs
```diff
@@ -21,8 +21,12 @@ const DUMMY_AMOUNT: u128 = 1;
 const BUFFER_TIME_SECONDS: u64 = 30;
 
 async fn sleep_until_new_timestamp() {
-    // Sometimes sleeping for 1 second isn't enough.
-    tokio::time::sleep(time::Duration::from_millis(1500)).await
+    let now = get_unix_timestamp_as_seconds();
+    tokio::time::sleep(time::Duration::from_millis(100)).await;
+
+    while now == get_unix_timestamp_as_seconds() {
+        tokio::time::sleep(time::Duration::from_millis(100)).await;
```
marioiordanov, line 28: this while loop can run up to 10 times, probably. I suggest increasing the Duration

3alpha: So, if we assume uniform distribution, there will be on average `duration / 2` of unnecessary waiting. I kept duration fairly short because of that, but we can increase. IDK, 250 ms?

marioiordanov: Before 1500ms waiting time was not enough, so here greater than 100ms IMO is better. I dont know what number exactly to be used

marioiordanov: 1 alternative is to check the amount of milliseconds that are passed in the current second and then wait enough milliseconds + buffer milliseconds to be sure that when called get_unix_timestamp_as_seconds will return different number. But as written in the referenced issue 1500ms didn't produce the expected behavior

3alpha: Given that this is just a test and doesn't influence devnet performance in the end, I would just keep it simple.

FabijanC: My fear is that there is a bug in the production code, not in the test code. I'd also like to add that the referenced issue #687 hasn't appeared in weeks, although when it did appear, it would usually be about once a week and in a workflow with no other failing tests (not entirely sure about that); and most of our CI workflows in the past months were not failure-free.

PR #755, opened 2025-04-11
0.8.1 JSON-RPC spec

marioiordanov:
## Usage related changes

Compliant with 0.8.1 spec

- each string component from JSON-RPC schema that doesn't have pattern, is assigned the default regex pattern (string of 1-10 characters )
- remove server dependency in tests/integration crate

## Checklist:

- [ ] Checked out the [contribution guidelines](https://github.com/0xSpaceShard/starknet-devnet/blob/main/.github/CONTRIBUTING.md)
- [ ] Linked the [issues](https://github.com/0xSpaceShard/starknet-devnet/issues) resolvable by this PR - [linking info](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword)
- [ ] Updated the tests if needed; all passing - [execution info](https://github.com/0xSpaceShard/starknet-devnet/blob/main/.github/CONTRIBUTING.md#test-execution)

## Summary by CodeRabbit

- **New Features**
  - Added enhanced validation rules for subscription IDs and contract class ABI parameters.
  - Introduced updated API specifications for accessing compiled contract artifacts.

- **API Updates**
  - Upgraded the API version from 0.8.0 to 0.8.1 across multiple endpoints.
  - Improved subscription identifier handling for increased consistency and clarity.

- **Internal Improvements**
  - Streamlined data generation with default patterns.
  - Consolidated utilities and removed redundant dependencies to simplify the testing environment.


crates/starknet-devnet-server/test_data/spec/0.8.1/edit_spec_instructions.yaml
```diff
@@ -27,6 +27,10 @@ add:
     new_entry: $ref/#/components/schemas/u64
   - path: components/schemas/CellRef/properties/offset
     new_entry: maximum/32767
+  - path: components/schemas/SUBSCRIPTION_ID
+    new_entry: pattern/^[1-9]{1}[0-9]{0,4}$
```
FabijanC, line 31: Why 1-5 chars? And why the first `{1}`?

crates/starknet-devnet-server/src/api/json_rpc/models.rs
```diff
@@ -201,7 +201,38 @@ pub struct L1TransactionHashInput {
     pub transaction_hash: Hash256,
 }
 
-pub type SubscriptionId = u64;
+// pub type SubscriptionId = u64;
```
FabijanC, line 204: Remnant

crates/starknet-devnet-server/src/api/json_rpc/models.rs
```diff
@@ -201,7 +201,38 @@ pub struct L1TransactionHashInput {
...
+    }
+}
+
+impl<'de> Deserialize<'de> for SubscriptionId {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        let u64_as_string = String::deserialize(deserializer)?;
+        let subscription_id = u64_as_string.parse::<u64>().map_err(|_| {
```
FabijanC, line 229: But does it really need to be a stringified u64 value? Can it not be any string value?

crates/starknet-devnet-server/src/api/json_rpc/spec_reader/data_generator.rs
```diff
@@ -14,6 +14,8 @@ use super::spec_schemas::tuple_schema::Tuple;
 use super::spec_schemas::{Primitive, Schema};
 
 const MAX_DEPTH: u8 = 5;
+// generate a random string with 1-10 characters
```
FabijanC, line 17: Perhaps we could use an explanation on which spec objects will use this (i.e. in what sense is it default).

Or, since this is the description of a constant, the "generate a random" part of the description does not belong here (it's more suitable for describing the piece of code where the constant is used).

And if is really the comment description of a const, it should use `///` instead o `//`.

crates/starknet-devnet-server/src/api/json_rpc/models.rs
```diff
@@ -201,7 +201,38 @@ pub struct L1TransactionHashInput {
     pub transaction_hash: Hash256,
 }
 
-pub type SubscriptionId = u64;
+// pub type SubscriptionId = u64;
+#[derive(Clone, Copy, Debug, Eq, PartialEq, PartialOrd, Ord, Hash)]
+pub struct SubscriptionId(u64);
```
FabijanC, line 206: Why is this better than `pub type SubscriptionId = String` ?

FabijanC (changes requested):

crates/starknet-devnet-server/test_data/spec/0.8.1/edit_spec_instructions.yaml
marioiordanov: I dont understand the question

FabijanC: There are two questions:
1) Why is the pattern satisfied if the length of the string is between 1 and 5 digits, inclusive?
2) Why does the pattern use `{1}`? Isn't it redundant?

crates/starknet-devnet-server/src/api/json_rpc/models.rs
marioiordanov: - better readability than string when debugging the actual ids
- generating a random string can include characters that can be not processed by clients
- older version was using integer. The change from integer to string is not to support wider range of characters, but due to a limitation in starknet.js to handle very big numbers
- from this discussion https://spaceshard.slack.com/archives/C03QN20522D/p1742228667777279 all the representative of the nodes mention that they use numbers
- It doesnt need a change in already working implementation of the ws functionality

crates/starknet-devnet-server/src/api/json_rpc/models.rs
marioiordanov: any string cant be parsed to u64 and the intention is to use u64 as the type for SubscriptionId

FabijanC: This relies on the assumption that, when the subscriber was originally subscribing, Devnet generated a stringified u64 value. This could be mentioned in a short comment.

crates/starknet-devnet-server/src/api/json_rpc/models.rs
FabijanC: Why does `PartialOrd` and `Ord` need to be derived?

crates/starknet-devnet-server/src/api/json_rpc/spec_reader/data_generator.rs
marioiordanov: https://github.com/0xSpaceShard/starknet-devnet/blob/70e6363f98dc0eee477651269c30e8f24485d4f2/crates/starknet-devnet-server/src/api/json_rpc/spec_reader/data_generator.rs

crates/starknet-devnet-server/src/api/json_rpc/models.rs
marioiordanov: It was a misconfusion that for retrieving values from the hashmap it needs to be able to order the keys. It works without these traits https://github.com/0xSpaceShard/starknet-devnet/blob/02e1b98a046ef7a04934a612d15e34d13f1bdbd2/crates/starknet-devnet-server/src/api/json_rpc/models.rs

crates/starknet-devnet-server/src/api/json_rpc/models.rs
marioiordanov: https://github.com/0xSpaceShard/starknet-devnet/blob/17a9c860714a6e8cde53a5e2ae8a5c81402b589f/crates/starknet-devnet-server/src/api/json_rpc/models.rs

crates/starknet-devnet-server/test_data/spec/0.8.1/edit_spec_instructions.yaml
marioiordanov: 2. I dont think integer can start with leading 0

FabijanC: Right, I was just saying that patterns `[1-9]` and `[1-9]{1}` are the same, but this is such a minute detail that I really regret pointing it out. We can keep it as it is.

marioiordanov: Now I understand what the comment is about [7329442](https://github.com/0xSpaceShard/starknet-devnet/pull/755/commits/7329442d9e69e71fa5ee30ffc9d7c4a3fff63b2d)

FabijanC: If length of 5 was chosen with no particular reason, note that for future readers of the config file.

marioiordanov: [718ca6e](https://github.com/0xSpaceShard/starknet-devnet/pull/755/commits/718ca6ea7a141d7275335d21f2443b388bdf0b2f)

FabijanC (approved):

tests/integration/common/utils.rs
```diff
@@ -9,6 +9,7 @@ use ethers::types::U256;
 use futures::{SinkExt, StreamExt, TryStreamExt};
 use rand::{Rng, thread_rng};
 use serde_json::json;
+use server::api::json_rpc::models::SubscriptionId;
```
FabijanC, line 12: I try to minimize the amount of code imported from production crates in this integration tests crate

marioiordanov: makes sense, why does the server crate is imported as a dependency?

FabijanC: Only members of `server::test_utils` are needed.

FabijanC: In fact, it seems only `assert_contains` is ever imported. Perhaps we could duplicate that simple function and have it also defined in tests/integration

marioiordanov: [5be8607](https://github.com/0xSpaceShard/starknet-devnet/pull/755/commits/5be8607c892a7e590675997a1b7a367381b23c12)

FabijanC (approved):

PR #801, opened 2025-06-20
Fix broken link to predeployed.md in account-impersonation.md

VolodymyrBg:
Updated the link to the predeployed accounts documentation in account-impersonation.md by adding the .md extension, ensuring the reference works correctly in the documentation site.


FabijanC:
It already works without the extension. If you convince me that it's necessary, I'll reopen the PR.

PR #959, opened 2026-08-12
Improve agent development workflow

adrienlacombe:
## Summary

- add a repository-wide agent guide and executable shell scripts for environment checks, routine verification, full validation, UI checks, website checks, and integration tests
- keep existing focused formatter, Clippy, spelling, and dependency scripts as the source of validation policy
- update contributor guidance, PR checklist, and CI to use the shell scripts or their exact underlying commands; no Makefile or task-runner dependency is added

## Validation

- ./scripts/doctor.sh
- bash -n for all new and existing validation scripts
- npm --prefix website ci && ./scripts/format_check.sh
- ./scripts/check_website.sh
- ./scripts/check_web_ui.sh
- workflow YAML parsing and git diff --check
- ./scripts/verify.sh: passed formatting and web validation, then stopped at the all-features Rust check because the local Homebrew toolchain cannot link libzstd

The previously run unit-test suite passed. Integration tests were not rerun because this change only reorganizes development tooling and CI commands.


.github/workflows/build-and-test.yml
```diff
@@ -2,6 +2,8 @@ name: Build and Test
 
 on:
   push:
+    branches: [main]
+  pull_request:
```
3alpha: Please revert old trigger.

.github/pull_request_template.md
```diff
@@ -10,14 +10,16 @@
...
+- [ ] No routine validation errors - `./scripts/verify.sh`
+- [ ] No unused dependencies - `./scripts/check_unused_deps.sh` (when Rust manifests changed)
 - [ ] No spelling errors - `./scripts/check_spelling.sh`
+- [ ] Ran integration tests - `./scripts/test_integration.sh` (when server, RPC, CLI, contracts, or integration tests changed)
+- [ ] Regenerated packaged UI assets - `npm --prefix web-ui run build:devnet` (when `web-ui` changed)
 - [ ] Performed code self-review
 - [ ] Rebased to the latest commit of the target branch (or merged it into my branch)
     -   Once you make the PR reviewable, please avoid force-pushing
-- [ ] Updated the docs if needed - `./website/README.md`
+- [ ] Updated the docs if needed - `npm --prefix website ci && ./scripts/check_website.sh`
```
3alpha: Revert this change, updating docs is not just running checks, but adding content as well if needed.

AGENTS.md
```diff
@@ -0,0 +1,93 @@
...
+`./scripts/doctor.sh` only warns when LLVM is missing and still exits 0.
+`./scripts/ci.sh` also runs spelling and integration tests.
+
+Use the focused scripts in `scripts/` for individual checks. State which targets you ran in
+the final handoff and name any skipped target with its reason.
+
+## Toolchain
+
+- Rust stable is pinned in `rust-toolchain.toml` (currently 1.96.0).
+- Formatting and spelling use `nightly-2026-04-10`; install it with
```
3alpha: Maybe it would be better if these toolchains are not hard-coded here, but introduced as reference?

3alpha (changes requested):
Addressed just a few details. Also, please do not do manual line breaks. You can add into `AGENTS.md` something like this:


```
### Markdown line wrapping

Never hard-wrap Markdown prose. A paragraph must remain on a single source line, regardless of its length. Do not insert newlines at 80, 100, 120, or any other column width.

Newlines are allowed only when they are semantically required by Markdown structure (e.g. headings, lists, blockquotes, code blocks, tables, or paragraph separation).
```

.github/workflows/build-and-test.yml
adrienlacombe: Restored the original workflow trigger block in commit 3c62ac60.

.github/pull_request_template.md
adrienlacombe: Restored the documentation checklist instruction in commit 3c62ac60.

AGENTS.md
adrienlacombe: Replaced hard-coded toolchain versions with configuration references and added the Markdown line-wrapping rule in commit 3c62ac60.

3alpha (approved):

